Issue #3424, opened 2014-01-18
Backspacing in share widget does not highlight, does not delete in bootstrap branch


mrvisser:
When using backspace in the share widget text-field, I expect it to:
1. First backspace, highlight the user behind my cursor
2. Second backspace, delete the highlighted item

What I'm seeing is:
- On master: First backspace does not highlight any of the items in the share text-field, second backspace deletes the item I would expect to be highlighted. When I click the X on an item, it does delete it.
- On Bootstrap3: First backspace does not highlight any of the items in the share text-field, second backspace does not delete the item. When I click the X on the item, it does not delete an item

This might be a regression related to: https://github.com/oaeproject/3akai-ux/pull/3365

PR #3699, opened 2014-04-01
Clickover plugin not compatible with bootstrap 3

bp323:
The [clickover](https://github.com/lecar-red/bootstrapx-clickover) plugin has not been updated in a year and is not compatible with bootstrap 3. (See, e.g. #3424)


nicolaasmatthijs:
Closing as the popover component is still working as expected on the latest bootstrap release

PR #3890, opened 2014-10-14
Allow entering of email addresses in autosuggest

sathomas:
- [ ] It should be possible to add an arbitrary email address to the autosuggest component. This should only be allowed when a valid email is entered and the user hits enter. This should just generate an autosuggest item that contains the email address.
- [ ] The placeholder text should read "Enter people, groups and email addresses"


nicolaasmatthijs:
@mrvisser : This is the UI ticket you'll need to incorporate once you get to inviting guests

PR #3895, opened 2014-10-20
Add reprocessing filter that targets specific tenant

sathomas:
The backend now allows you to [reprocess content for a specific tenant](https://github.com/oaeproject/Hilary/pull/1050). It would be good to have this in the admin UI as well.

I'm not sure how much work it'd be, but it'd be cool if the filters were combinable as they are in the backend.

For example, say a tenant gets a new video processing thingie, it'd be cool if we could select the tenant in a dropdown and then select the appropriate mime types to reprocess for that tenant. That can be outscoped from this ticket though.

---
This change is [Reviewable](https://reviewable.io/reviews/oaeproject/3akai-ux/3895)


nicolaasmatthijs:
@sathomas : Looks like there is a merge conflict here

sathomas:
Merge de-conflicted. To @nicolaasmatthijs for review.

PR #3908, opened 2014-12-11
Etherpad 1.5.6 work

simong:


nicolaasmatthijs:
@simong : Is this PR still valid?

simong:
Yes, it's necessary for 1.5.6 as well. I'll rename the issue's title

mrvisser:
@simong want me to review this for the bug bash?

nicolaasmatthijs:
Sounds like that will be useful. Over to @mrvisser

mrvisser:
@simong can you give an overview of why this change is being made? Also it will need a puppet-hilary PR to go along w/ it so we can pull it into the bug bash.

simong:
Etherpad upgraded their version of socket.io which changed where the websocket is exposed.

IIRC, by default it will be exposed at `http://domain/socket.io` but we've [contributed a patch](https://github.com/ether/etherpad-lite/commit/45c1ad078bda7800cec1a81f2b8517adec93c43d) that exposes it at `http://baseURL/socket.io`. Because we embed etherpad using the URL: `http://tenantHost/etherpad/0` that baseURL will be `tenantHost/etherpad/0`.

I'll have a puppet-hilary branch ready that I'll push in a minute.

PR #3909, opened 2014-12-16
(A11y severity 1) Preserve keyboard focus on file uploads

sathomas:
When a user selects a file to upload in the "Upload file(s)" modal, the modal changes causing keyboard focus to be lost. Once keyboard focus is lost, the user essentially becomes trapped and cannot navigate at all. Ensure that keyboard focus is set to a logical place within the modal once a file has been selected.

### DIFF
--- a/node_modules/oae-core/upload/js/upload.js
+++ b/node_modules/oae-core/upload/js/upload.js
@@ -399,7 +399,6 @@ define(['jquery', 'oae.core', 'jquery.fileupload', 'jquery.iframe-transport', 'j
                 'url': '/api/content/create',
                 'dropZone': $('#upload-dropzone', $rootel),
                 'forceIframeTransport': useIframeTransport,
-                'replaceFileInput': false,
                 // Drop is fired when a user drops files on the dropzone
                 'drop': function(ev, data) {
                     // Ensure at least one file is valid


nicolaasmatthijs:
@bp323 Would you be able to provide some insight here?

bp323:
Me memory is letting me down on the specifics but I can give you some pointers to further test this. By setting `replaceFileInput: false` it will not swap out the file input. This can have as a side effect that selecting the same file again might not trigger the change event on that input, which the upload widget relies on to build the list of files.

If anything funny is going to happen with uploads it'll be on IE so I'd double check this in there. I'm not sure if the uploadnewversion widget is using this setting but maybe there's more clarification to be found in there as well.

The FAQ of jquery.fileupload might hold some answers as well (https://github.com/blueimp/jQuery-File-Upload/wiki/Frequently-Asked-Questions)

nicolaasmatthijs:
@sathomas : Can you verify that this is working as expected on all browsers, including IE? That should include browsing and selecting a single file, browsing and selecting multiple files and drag and dropping files.

sathomas:
Working from home today so I can check this on IE. I was preparing to resurrect my IE9 testing tools, but I just checked Google Analytics and the production site only had 12 visitors using IE9 in the past four weeks. @nicolaasmatthijs: do we still need to support IE9? (IE10 and later support XHR file uploads, so the option above is really only relevant for IE9).

sathomas:
Works about as best as it can. Details:

* Uploading a single file using keyboard
    - [x] Internet Explorer 11
    - [x] Safari
    - [x] Chrome
    - [x] Firefox
    - [x] Vivalidi
* Uploading multiple files using keyboard
    - [ ] **Internet Explorer 11 Does not work** After closing the file open dialog, Windows returns focus to the IE chrome rather than the content of the page. There's no way to get to the dialog using the keyboard, so the user cannot upload the files. This is a Windows/IE issue, however, as there is nothing we can do to fix it. (Workaround for users is to upload one file at a time.)
    - [x] Safari
    - [x] Chrome
    - [x] Firefox
    - [x] Vivalidi
* Uploading multiple files using drag and drop
    - [x] Internet Explorer 11
    - [x] Safari
    - [x] Chrome
    - [x] Firefox
    - [x] Vivalidi

> Note: I am unable to test with versions of IE earlier than 11. Ironically, there seems to be a conflict between VirtualBox and the Logitech keyboard driver which makes the keyboard unusable once VirtualBox starts a VM. The testing above was using IE11 on real hardware.

I also tested various combinations of selecting files, canceling them, and then re-selecting them for upload. All seemed to work as expected.

Not sure how much "review" there is for a one-line deletion, but I'll pass this to @nicolaasmatthijs for next steps.

node_modules/oae-core/upload/js/upload.js
```diff
@@ -399,7 +399,6 @@ define(['jquery', 'oae.core', 'jquery.fileupload', 'jquery.iframe-transport', 'j
                 'url': '/api/content/create',
                 'dropZone': $('#upload-dropzone', $rootel),
                 'forceIframeTransport': useIframeTransport,
-                'replaceFileInput': false,
```
nicolaasmatthijs: @sathomas : What should we do with this in the other widgets it's being used in:

```
node_modules/oae-admin/importusers/js/importusers.js
node_modules/oae-core/uploadnewversion/js/uploadnewversion.js
```

nicolaasmatthijs:
Left a comment. Back to @sathomas

node_modules/oae-core/upload/js/upload.js
sathomas: It's not a problem on uploadnewversion because that widget doesn't present an intermediate dialog for confirmation. It simply uploads the new file immediately. It's also not a problem with importusers. I'm not sure exactly why, but the form control layout is significantly different and that's apparently enough to prevent triggering the Firefox bug.

We could make the changes in those widgets as well for consistency, but from a functional viewpoint they aren't required.

nicolaasmatthijs:
Looks good. Merged.

PR #3914, opened 2014-12-30
(A11y severity 2) Associate notifications dropdown content more closely with icon

sathomas:
The "Notifications" icon (), when selected, opens a drop down of recent notifications. While the drop down is accessible by keyboard, users can only access it after they have navigated through the entire page. Keyboard focus should be set to the dropdown content upon selecting the icon or, optimally, the underlying code should be rearranged to ensure the dropdown content comes immediately after the icon instead of at the end of the page.

### DIFF
--- a/node_modules/oae-core/topnavigation/topnavigation.html
+++ b/node_modules/oae-core/topnavigation/topnavigation.html
@@ -155,7 +155,7 @@
                 <i class="fa fa-home"><span class="sr-only">__MSG__HOME__</span></i>
             </a>
         </li>
-        <li id="menuitem-notifications">
+        <li id="topnavigation-notifications-container">
             <button type="button" class="btn btn-link oae-trigger-notifications" title="__MSG__NOTIFICATIONS__" role="menuitem">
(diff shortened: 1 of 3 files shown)


node_modules/oae-core/topnavigation/topnavigation.html
```diff
@@ -155,7 +155,7 @@ <h1 id="topnavigation-target" class="sr-only">__MSG__SKIP_MENU_TARGET__</h1>
                 <i class="fa fa-home"><span class="sr-only">__MSG__HOME__</span></i>
             </a>
         </li>
-        <li>
+        <li id="menuitem-notifications">
```
nicolaasmatthijs: Needs to be namespaced with `topnavigation-`. Something like `topnavigation-notifications-container` would be consistent with the codebase.

nicolaasmatthijs:
Back to @sathomas for minor follow-up

sathomas:
followed up on review. To @nicolaasmatthijs to check

nicolaasmatthijs:
Looks good. Merged.

PR #3918, opened 2014-12-31
(A11y severity 2) Make visibility icons accessible to screen readers

sathomas:
All of the visibility icons are icon fonts. When encountered by a screen reader, the underlying text character is not read; however, since the icon presents information that is important to the user there needs to be a way to indicate what privacy level each document has. Hide the character with aria-hidden="true" and add off-screen text that presents the content (e.g., <span class="fa fa-oae-private" aria-hidden="true"> #</span>). Off-screen content should be styled so it remains accessible to screen reader users as documented at http://webaim.org/techniques/css/invisiblecontent/


nicolaasmatthijs:
Before I do a detailed review, is this really going to be an overall accessibility improvement? Rather than being able to quickly go through a list of items, people on screenreaders will now have to go through a full sentence of visibility information for each list item, which seems like it would make it a lot slower and a lot more difficult. The visibility icons were always designed as a hint, rather than something you'd have to go through to have a full experience.

Wouldn't it be more important to ensure the visibility icons in the clip thumbnails on the content/group/discussion profile pages have some screenreader only text?

sathomas:
@nicolaasmatthijs Your analysis makes a lot of sense to me. I think the issue is that strict WCAG 2.0 compliance from [Guideline 1.1](http://www.w3.org/TR/WCAG20/#text-equiv):

> All non-text content that is presented to the user has a text alternative that serves the equivalent purpose

There is an exception for pure decoration, but it would seem like a stretch to classify the icons as [purely decorative](http://www.w3.org/TR/WCAG20/#puredecdef):

> serving only an aesthetic purpose, providing no information, and having no functionality

In general, I'd be in favor of making the best decision for the users and not worrying so much about strict standards compliance. Since I have so little direct experience, though, I wonder if visually impaired folks are used to getting a lot of "extra" content read to them on web pages, in which case having the visibility read to them in a list wouldn't be that much of a bother.

I'll try both modes with Voiceover. It won't really be a valid comparison since I'm not a frequent Voiceover user, but it will, at least satisfy my curiosity.

nicolaasmatthijs:
@sathomas: We can indeed not argue that the icons are purely decorative, as they do convey some meaning. It might indeed be true that screenreader users are very used to moving through lots of extra content. Would you be able to follow up with Jared to check if that's the case?

If this would block us from being certified, we should probably implement it regardless.

sathomas:
@nicolaasmatthijs: I did spend some time experimenting with VoiceOver today. (@stuartf is no doubt glad I'm working from home. Guess I'll have to break out the headphones when I'm in the office.) At least with VoiceOver, I think it's pretty safe to go ahead and include the screen reader text with all icons. Here are the details:

* On initial load, VO doesn't read all of the checkbox items. It only reads a checkbox item when you tab to it.
* Furthermore, when you tab to a checkbox, VO only reads the `<label>` associated with the checkbox. It does not read the meta-data that includes the visibility icon.
* If you tab to a checkbox item and then use the arrow keys, it is possible to access the meta-data, but VO only reads a character at a time (including, e.g., the title and file name for content in a library). In this mode, it will access screen reader content for the visibility icons, but it's hard to imagine that this type of interaction would be useful in any case (not only is it painful, but the label already conveys the most relevant information—the title—anyway), so I don't think there's any harm in including screen reader text.
* The most obvious way to access the meta-data is using the Safari Web Rotor. There are several ways to use WR to "view" the meta-data, but they all work pretty similar to the most direct way:
  * WR can provide a list of links on the page that you can arrow-key through. Since the meta-data is enclosed in `<a>` tags, this list is effectively a list of links with their meta-data.
  * VO only reads the meta-data for a link when you "scroll" to it using the arrow keys, so you don't have to listen to any extraneous content when you invoke WR.
  * If you're "scrolling" through the list using the arrow keys, VO reads the title before it reads the visibility information. As soon as you hear the (start of the) title, you can immediately arrow-key to the next item without bothering to listen to the visibility information.

Based on what I've observed so far, I think it's fine to go ahead and include the screen reader text in all instances of the visibility icon. I can ping the WebAIM folks to see if JAWS is significantly different than VoiceOver, but now that I've seen how VO works, I'd be surprised if JAWS didn't do much the same thing. The VO approach makes a lot of sense.

nicolaasmatthijs:
@sathomas : Okay, that sounds good. Thanks for this thorough analysis. I'll proceed with the review.

shared/oae/bundles/ui/default.properties
```diff
@@ -498,6 +498,9 @@ USERNAME = Username
 USERNAME_COLON = Username:
 USER_MANAGEMENT = User management
 VIDEO = Video
+VISIBILE_ONLY_TO_SPECIFIED_USERS = Visible only to specified users
```
nicolaasmatthijs: I think it would be good if this could re-use the visibility messages from the manageaccess for the various entitites:

*Files*

```
FILE_PRIVATE_DESCRIPTION
FILE_LOGGEDIN_DESCRIPTION
FILE_PUBLIC_DESCRIPTION
```

*Collaborative documents*

```
DOCUMENT_PRIVATE_DESCRIPTION
DOCUMENT_LOGGEDIN_DESCRIPTION
DOCUMENT_PUBLIC_DESCRIPTION
```

*Links*

```
LINK_PRIVATE_DESCRIPTION
LINK_LOGGEDIN_DESCRIPTION
LINK_PUBLIC_DESCRIPTION
```

*Groups*

```
GROUP_PRIVATE_DESCRIPTION_PRESENT
GROUP_LOGGEDIN_DESCRIPTION
GROUP_PUBLIC_DESCRIPTION_PRESENT
```

*Discussions*

```
DISCUSSION_PRIVATE_DESCRIPTION
DISCUSSION_LOGGEDIN_DESCRIPTION
DISCUSSION_PUBLIC_DESCRIPTION
```

*Folders*

```
FOLDER_PRIVATE_DESCRIPTION
FOLDER_LOGGEDIN_DESCRIPTION
FOLDER_PUBLIC_DESCRIPTION
```

nicolaasmatthijs: Note that we have to make sure that the tenant displayName replacement in the `LOGGEDIN` descriptions should use the entity tenant and not the current user's tenant

shared/oae/macros/list.html
```diff
@@ -360,3 +360,22 @@ <h3 class="oae-threedots"><span class="jeditable-field oae-threedots" rel="toolt
         <label class="sr-only" for="oae-list-${type}-${entityData.id}">${entityData.displayName|encodeForHTML}</label>
     {/if}
 {/macro}
+
+<!--
+ * Render a visibility icon with support for screen readers
+ *
+ * @param  {String}     visibility      ["public" | "loggedin" | "private"]
+-->
+{macro renderVisibilityIcon(visibility)}
```
nicolaasmatthijs: The comment above means you'll need to pass in the full entity

nicolaasmatthijs:

sathomas:
using existing description messages based on content. To @nicolaasmatthijs for review

shared/oae/macros/list.html
```diff
@@ -360,3 +361,91 @@ <h3 class="oae-threedots"><span class="jeditable-field oae-threedots" rel="toolt
         <label class="sr-only" for="oae-list-${type}-${entityData.id}">${entityData.displayName|encodeForHTML}</label>
     {/if}
 {/macro}
+
+<!--
+ * Render a visibility icon with support for screen readers
+ *
+ * @param  {User|Group|Content}     entityData     Standard object representing a user, group or content item or a search result for a user, group or content item as returned by Hilary
```
nicolaasmatthijs: Should have `Folder` and `Discussion` as well

nicolaasmatthijs: Can you update the docs for `listItem` as well?

nicolaasmatthijs:
Added some additional review. Back to @sathomas

sathomas:
good catch @nicolaasmatthijs docs updated now

shared/oae/macros/list.html
```diff
@@ -105,15 +105,15 @@
...
- * @param  {String}                 [displayOptions.customImage]           URL to the custom entity profile picture that should be used. If not provided, this will fall back to the thumbnailUrl
- * @param  {Function}               [displayOptions.listItemActions]       Macro that renders list item actions
- * @param  {String}                 [displayOptions.additionalClasses]     Additional CSS classes separated by spaces to apply to the generated list items
- * @param  {Boolean}                [displayOptions.addLink]               Whether or not a link to the entity profile should be placed on the list item. By default, a link will be set
- * @param  {String}                 [displayOptions.linkTarget]            The target of the list item link. One of `_blank`, `_self`, `_parent` or `_top`. By default, the link will open in the same window
+ * @param  {User|Group|Content|Discussion|Folder}   entityData                             Standard object representing a user, group or content item or a search result for a user, group or content item as returned by Hilary
```
nicolaasmatthijs: There were a few other places that needed this update.

nicolaasmatthijs:
Followed up and merged

PR #3919, opened 2015-01-02
(A11y severity 2) Add grouping and legend to form controls (or equivalent)

sathomas:
Checkboxes and radio buttons are not properly associated with their labels or descriptions, as, for example, on the search form.

Users navigating through the page might be confused when reading the checkboxes as the "Refine your search" label for the boxes is not associated with them. This can be resolved in multiple ways:

* Place the heading and all associated options within a `<fieldset>` element with the heading being a `<legend>` for that fieldset. See http://webaim.org/techniques/forms/controls#checkbox

* Mark up radio buttons and checkboxes within an element with `role="radio group"`. This will identify the answers as a group. Then use `aria-describedby` on this element to reference the element that contains the descriptive text. This will cause the screen reader to read the question text for each answer within the radiogroup (i.e., identical functionality to fieldset/legend).

* Add `tabindex="0"` to the label itself, thus placing it in the keyboard navigation flow and causing it to be navigated to and read previous to the privacy options.

The privacy radio buttons on file upload modals have a similar problem

### DIFF
--- a/node_modules/oae-core/setpermissions/setpermissions.html
+++ b/node_modules/oae-core/setpermissions/setpermissions.html
@@ -5,31 +5,35 @@
 
 <div id="setpermissions-template"><!--
     <div id="setpermissions-form" role="form">
-        <h4>__MSG__VISIBILITY__</h4>
-        <div class="radio">
-            <label>
-                <input type="radio" value="private" name="setpermissions-visibility" class="pull-left" {if visibility === 'private'} checked="checked"{/if}/>
-                <i class="fa fa-oae-private pull-left"></i>
-                <span>__MSG__PRIVATE__</span>
-            </label>
-        </div>
-        <div class="radio">
-            <label>
-                <input type="radio" value="loggedin" name="setpermissions-visibility" class="pull-left" {if visibility === 'loggedin'} checked="checked"{/if}/>
-                <i class="fa fa-oae-loggedin pull-left"></i>
-                <span>${oae.data.me.tenant.displayName|encodeForHTML}</span>
-            </label>
-        </div>
-        <div class="radio">
-            <label>
-                <input type="radio" value="public" name="setpermissions-visibility" class="pull-left" {if visibility === 'public'} checked="checked"{/if}/>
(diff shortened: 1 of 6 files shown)


sathomas:
This fixes all the issues in the user interface (complements of a tricky workaround for a webkit bug).

There are two places in the admin interface that have a potential problem: reprocess specific content types and reprocess specific mime types. Fixing those instances will require a bit of a different approach since they're not using the same HTML structure. I'm not sure of the priority for accessibility of the admin UI, but in case it's not as high a priority as the user UI, I haven't included fixes for the admin UI in this PR (yet).

To @nicolaasmatthijs for review and comment. In addition to review, let me know if you want to add the admin fixes in this PR or create a separate issue for them.

node_modules/oae-core/setpermissions/setpermissions.html
```diff
@@ -5,31 +5,35 @@
...
-            </label>
-        </div>
-        <div class="radio">
-            <label>
-                <input type="radio" value="public" name="setpermissions-visibility" class="pull-left" {if visibility === 'public'} checked="checked"{/if}/>
-                <i class="fa fa-oae-public pull-left"></i>
-                <span>__MSG__PUBLIC__</span>
-            </label>
-        </div>
+        <fieldset>
```
nicolaasmatthijs: @sathomas : What about the modals where we have the large list options (creategroup, createcollabdoc, manageaccess, etc.). Do the visibility options there also need to be changed to a fieldset or not?

nicolaasmatthijs:
Left a question for @sathomas

node_modules/oae-core/setpermissions/setpermissions.html
sathomas: Yes they do. (I missed those because I was grepping for the bootstrap class instead of the input type.) Alas, they're going to be a bit more challenging.

sathomas:
I think this covers all the grouped radio buttons and checkboxes; I fixed the admin UI as well since it wasn't as complex as I first thought. To @nicolaasmatthijs for review

PR #3922, opened 2015-01-06
(A11y severity 2) Remove redundant links from grid contents

sathomas:
User name and user images are another common issue throughout the site with redundant information and links being presented to users.
The image container as well as the name "Jonathan Whiting" both link to the same location. Additionally, the image container has an aria-label that matches the name text. This causes the name to be repeated unnecessarily by screen readers (this is particularly problematic and cumbersome when comments are read within the activity stream). Since the context of the image can easily be understood by the following name, remove the ARIA label, and either remove the link from the image container (the element could be scripted to respond to mouse click) or combine the image and the adjacent text ("Added by Jonathan Whiting") into one link.

### DIFF
--- a/node_modules/oae-core/activity/activity.html
+++ b/node_modules/oae-core/activity/activity.html
@@ -61,7 +61,7 @@
                             <div class="media-body">
                                 <h4 class="media-heading">
                                     {if item.comment.author['oae:profilePath']}
-                                        <a href="${item.comment.author['oae:profilePath']|profilePath}">${item.comment.author.displayName|encodeForHTML}</a>
+                                        <a href="${item.comment.author['oae:profilePath']|profilePath}" aria-hidden="true">${item.comment.author.displayName|encodeForHTML}</a>
                                     {else}
                                         ${item.comment.author.displayName|encodeForHTML}
                                     {/if}
--- a/node_modules/oae-core/comments/comments.html
+++ b/node_modules/oae-core/comments/comments.html
@@ -52,7 +52,7 @@
                     </div>
                     <h4 class="media-heading">
                         {if comment.createdBy.profilePath}
-                            <a href="${comment.createdBy.profilePath|profilePath}">
+                            <a href="${comment.createdBy.profilePath|profilePath}" aria-hidden="true">
                                 ${comment.createdBy.displayName|encodeForHTML}
                             </a>
                         {else}
--- a/shared/oae/macros/list.html
+++ b/shared/oae/macros/list.html
@@ -268,7 +268,7 @@
     <div class="oae-thumbnail {if displayOptions.largeDefault} oae-thumbnail-large {/if} fa fa-oae-${resourceSubType} fa-oae-${resourceType}" data-id="${resourceId}">
         {if profilePath && displayOptions.addLink !== false}<a href="${profilePath|profilePath}" title="${entityData.displayName|encodeForHTMLAttribute}" {if displayOptions.linkTarget} target="${displayOptions.linkTarget}"{/if}>{/if}
         {if thumbnailUrl}
-            <div role="img" aria-label="${entityData.displayName|encodeForHTMLAttribute}" style="background-image: url('${thumbnailUrl}');"></div>
+            <div role="img" style="background-image: url('${thumbnailUrl}');"></div>
         {/if}
         {if displayOptions.addVisibilityIcon !== false && resourceType !== 'user'}
             ${renderVisibilityIcon(entityData)}


sathomas:
To @nicolaasmatthijs for review. Note that I did not remove the link from the thumbnail as suggested; not only would that make the markup less semantic, but we use thumbnails all over the place and it would be error-prone to try and distinguish those uses which have a nearby text link from those that don't. Instead, I added the `aria-hidden` attribute to the text version of the link. Although WCAG strongly discourages using `aria-hidden` with content that is otherwise visible to sighted users, there is [an exception](http://www.w3.org/TR/wai-aria/states_and_properties#aria-hidden):

> Authors **MAY,** with caution, use aria-hidden to hide visibly rendered content from assistive technologies _only_ if the act of hiding this content is intended to improve the experience for users of assistive technologies by removing redundant or extraneous content. Authors using aria-hidden to hide visible content from screen readers **MUST** ensure that identical or equivalent meaning and functionality is exposed to assistive technologies.

I think this covers our case precisely.

node_modules/oae-core/activity/activity.html
```diff
@@ -61,7 +61,7 @@
                             <div class="media-body">
                                 <h4 class="media-heading">
                                     {if item.comment.author['oae:profilePath']}
-                                        <a href="${item.comment.author['oae:profilePath']|profilePath}">${item.comment.author.displayName|encodeForHTML}</a>
+                                        <a href="${item.comment.author['oae:profilePath']|profilePath}" aria-hidden="true">${item.comment.author.displayName|encodeForHTML}</a>
```
nicolaasmatthijs: When I try this with VoiceOver enabled, tabbing through the comments still stops at both the thumbnail image and the user link, causing the name of the user to be read twice. Should we also be removing the tabindex from these thumbnail images to ensure that the name is not repeated?

nicolaasmatthijs:
Added a question. Back to @sathomas

sathomas:
Found it. There's a difference between activity summary and activity preview. Fix incoming.

sathomas:
Not as easy I as was thinking. The markup for the activity summary is actually generated by the server. It's the `actor1Link` property in the `activity` object. Now I'm not sure how practical it is to fix this specific case. It definitely doesn't seem like a good idea to modify the server code since that code can't be sure that the actor link will always be positioned next to a thumbnail. I'll have another look at hiding the thumbnail from screen readers instead of the text link.

The current PR is definitely an improvement. Before, screen readers were speaking each entity three times. With the PR screen readers speak entities one time unless the entity is in the activity summary, in which case it is spoken twice.

nicolaasmatthijs:
Note that the `actor1Link` is not generated by the server. It happens in `_setSummaryPropertiesForEntity` in the `activityadapter.js`

sathomas:
Decided to go with hiding the thumbnail link (using `aria-hidden`) instead of the text link. To @nicolaasmatthijs

nicolaasmatthijs:
@sathomas : I've gone through all of the places where we use `renderThumbnail` and couldn't find any places where we don't want the `aria-hidden` on the thumbnail, as the thumbnail is always accompanied by a link (including tiles). Maybe we can just add `aria-hidden` to all of them without adding it as a configuration option.

nicolaasmatthijs:

sathomas:
@nicolaasmatthijs we could certainly hardcode `aria-hidden` for all thumbnails, but I'd worry a bit about that. WCAG puts all kinds of warnings about using `aria-hidden` and they really want you to be sure that you use it only when you're certain that the content is available in some other element. Right now, we can say that all the cases in which we render a thumbnail, we also include a text link, but in the future that might not be true. A developer that wasn't familiar with accessibility might just use `renderThumbnail` without realizing the implications of hiding it from screen readers. By making this a parameter, we're at least forcing developers to think about the problem and make a conscious decision.

node_modules/oae-core/activity/activity.html
```diff
@@ -20,7 +20,8 @@
     {macro renderSummary(activity)}
         <div class="activity-summary-container clearfix">
             <div class="pull-left activity-entity-thumbnail">
-                ${renderThumbnail(activity.primaryActor)}
+                {var displayOptions = new Object(); displayOptions.hideLink = true}
+                ${renderThumbnail(activity.primaryActor, displayOptions)}
```
nicolaasmatthijs: Rather than creating the `displayOptions` object in the Trimpath template, can we pass it in through js? That's the pattern we're mostly using for displayOptions now

node_modules/oae-core/activity/activity.html
```diff
@@ -56,7 +57,8 @@
                     {for item in activity.latestComments}
                         <li class="media activity-level-${item.level}">
                             <div class="pull-left activity-entity-thumbnail">
-                                ${renderThumbnail(item.comment.author)}
+                                {var displayOptions = new Object(); displayOptions.hideLink = true}
```
nicolaasmatthijs: Same here

node_modules/oae-core/comments/comments.html
```diff
@@ -32,7 +32,8 @@
         {var commentLevel = comment.level > 2 ? 2 : comment.level}
         <li id="${comment.threadKey}" class="media comments-level-${commentLevel} {if !comment.body} deleted{/if}" data-id="${comment.created}">
             <div class="comments-thumbnail">
-                ${renderThumbnail(comment.createdBy || 'user')}
+                {var displayOptions = new Object(); displayOptions.hideLink = true}
```
nicolaasmatthijs: Same here

node_modules/oae-core/notifications/notifications.html
```diff
@@ -25,7 +25,8 @@
                 <div class="oae-list-details-item">
                     <div class="oae-listitem{if hasTargetThumbnail} notifications-with-target-thumbnail{/if}">
                         <div class="oae-listitem-primary-thumbnail">
-                            ${renderThumbnail(activity.primaryActor)}
+                            {var displayOptions = new Object(); displayOptions.hideLink = true}
```
nicolaasmatthijs: Same here

node_modules/oae-core/members/members.html
```diff
@@ -31,6 +31,7 @@
         {else}
             {var displayOptions = new Object()}
             ${displayOptions.metadata = role|eat}
+            ${displayOptions.hideLink = true|eat}
```
nicolaasmatthijs: Maybe we could do that here as well?

nicolaasmatthijs:

sathomas:
got rid of `displayOptions` in trimpath in favor of JavaScript. To @nicolaasmatthijs

nicolaasmatthijs:
This is looking good. However, I think we also need to add this to the following widgets:

- aboutcontent                     
- aboutdiscussion                                
- aboutfolder
- aboutgroup

nicolaasmatthijs:

sathomas:
Remembered some other instances as well; also cleaned up documentation a little bit. Back to @nicolaasmatthijs

nicolaasmatthijs:
Left some more feedback. Back to @sathomas for follow-up

sathomas:
now hiding thumbnail links in all lists. back to @nicolaasmatthijs for more review

nicolaasmatthijs:
Looks like there is a merge conflict here. Back to @sathomas for resolution.

sathomas:
merge de-conflicted. Back to @nicolaasmatthijs

PR #3923, opened 2015-01-07
(A11y severity 2) Eliminate repetition in activity stream (screen readers)

sathomas:
Throughout the recent activities page, content contains `<time>` elements that notify the user of how long ago an item was added to the activity stream (e.g., ). Every minute these elements update themselves and then are read by screen reader users. For example, a user who does nothing on the page and has nothing shared or added to their activity stream will hear something like this every minute: 
"13 minutes ago, 7 days ago, 14 minutes ago, 13 minutes ago, 34 minutes ago, 5 days ago, 7 days ago, 7 days ago, 7 days ago 7 days ago, 7 days ago, 7 days ago, 10 days ago, 7 days ago, 8 days ago, 8 days ago"
While the reason why screen readers continually read this information is unclear, it should be resolved by giving the `<time>` elements the ARIA attribute of `aria-live="off"` or by using a `<div>` instead of `<time>`.

### DIFF
--- a/node_modules/oae-core/activity/activity.html
+++ b/node_modules/oae-core/activity/activity.html
@@ -24,7 +24,7 @@
             </div>
             <div class="activity-summary">
                 ${oae.api.i18n.translate(activity.summary.i18nKey, 'activity', activity.summary.i18nArguments)}
-                <small class="muted"><time datetime="${activity.published}"></time></small>
+                <small class="muted"><time aria-live="off" datetime="${activity.published}"></time></small>
             </div>
         </div>
     {/macro}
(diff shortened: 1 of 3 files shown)


sathomas:
ready for @nicolaasmatthijs review

node_modules/oae-core/activity/activity.html
```diff
@@ -65,7 +65,7 @@ <h4 class="media-heading">
                                     {else}
                                         ${item.comment.author.displayName|encodeForHTML}
                                     {/if}
-                                    <small class="muted"><time datetime="${item.comment.published}"></time></small>
+                                    <small class="muted"><time aria-live="off" datetime="${item.comment.published}"></time></small>
```
nicolaasmatthijs: Rather than setting `aria-live="off"` in all `time` instances that use timago, we could just add it to `var timeAgo = exports.timeAgo = function($container) {` in the l10n api, just before `$timeEl.timeago();`

nicolaasmatthijs:

sathomas:
Definitely a better approach. Back to @nicolaasmatthijs

nicolaasmatthijs:
Looks good. Merged.

PR #3925, opened 2015-01-07
(A11y severity 2) Tooltip accessibility

sathomas:
When a user navigates to the file name after selecting it, a tooltip appears visually (Click to rename), but this information is not presented to screen reader users. This field may be associated to the element via `aria-describedby`.


sathomas:

nicolaasmatthijs:
Looks good. Merged.

PR #3926, opened 2015-01-08
(A11y severity 2) Label page number input in document preview

sathomas:
The page number text box does not have an associated label. Because there is no text next to this field that informs the user to the purpose of the textbox (e.g., "Enter page number "), the label will need to be hidden off-screen using CSS (see webaim.org/techniques/css/invisiblecontent/) or the field given a descriptive `title` or `aria-label` attribute.


sathomas:
To @nicolaasmatthijs for review. I think the `aria-label` attribute is even cleaner than adding a `.sr-only` `<label>`, but `sr-only` will work if you want to go that way instead.

nicolaasmatthijs:
This looks good. Merged.

PR #3928, opened 2015-01-08
(A11y severity 2) Limit alerts in activity stream

sathomas:
When a user’s activity stream is updated with new content, the new content on the page is given `role="alert"`. This causes all the new content on the page to be read to a screen reader, which is typically too much information. Additionally a consistent method for notifying users should be maintained. (cf. issue https://github.com/oaeproject/3akai-ux/issues/3921)

A much simpler and succinct approach would be to set the alert to the banner that appears when content is updated:

This method not only provides all pertinent information, but also avoids overwhelming users with information. Users would then be free to either navigate to the activity stream or continue their current task.

### DIFF
--- a/node_modules/oae-core/activity/js/activity.js
+++ b/node_modules/oae-core/activity/js/activity.js
@@ -122,8 +122,9 @@ define(['jquery', 'underscore', 'oae.core', 'activityadapter'], function($, _, o
             infinityScroll = $('.oae-list', $rootel).infiniteScroll(url, {
                 'limit': 10
             }, '#activity-items-template', {
-                'postProcessor': processActivities,
-                'emptyListProcessor': handleEmptyResultList
+                'ariaLive': 'off',  // Notifications will alert screen readers
+                'emptyListProcessor': handleEmptyResultList,
+                'postProcessor': processActivities
             });
         };
 
--- a/shared/oae/js/jquery-plugins/jquery.infinitescroll.js
+++ b/shared/oae/js/jquery-plugins/jquery.infinitescroll.js
@@ -34,6 +34,7 @@ define(['jquery', 'underscore', 'oae.api.util', 'oae.api.i18n', 'oae.api.l10n'],
      * @param  {Object}                            [options]                       Optional object containing additional configuraton options.
      * @param  {String|Element}                    [options.scrollContainer]       jQuery element or selector for that jQuery element that identifies the container on which the scrollposition should be watched to check when we are close enough to the bottom to load a new set of results. If this is not provided, the document body will be used.
      * @param  {String|Function}                   [options.initialContent]        HTML string that should be prepended to the list upon initialization. If a function is provided, the function will be called with no parameters and should return the HTML string to prepend.
+     * @param  {String         }                   [options.ariaLive]              `aria-live` attribute for container. Defaults to "assertive"
      * @param  {Function}                          [options.emptyListProcessor]    Function that will be executed when the rendered list doesn't have any elements.
      * @param  {Function}                          [options.postProcessor]         Function used to transform the search results before rendering the template. This function will be called with a data parameter containing the retrieved data and should return the processed data
      * @param  {Function}                          [options.postRenderer]          Function executed after the rendered HTML has been appended to the rendered list. The full retrieved server response will be passed into this function.
@@ -49,7 +50,9 @@ define(['jquery', 'underscore', 'oae.api.util', 'oae.api.i18n', 'oae.api.l10n'],
 
         // Default values
         parameters = parameters || {};
-        options = options || {};
+        options = $.extend({
+            'ariaLive': 'assertive'
+        }, options);
 
         // Number of items to load per call to the server
         parameters.limit = parameters.limit || 10;
@@ -386,7 +389,7 @@ define(['jquery', 'underscore', 'oae.api.util', 'oae.api.i18n', 'oae.api.l10n'],
         // Initialisation //
         ////////////////////
 
-        $listContainer.attr('aria-live', 'assertive');
+        $listContainer.attr('aria-live', options.ariaLive);
         setUpLoadingImage();
         setUpInitialContent();
         loadResultList();


sathomas:

node_modules/oae-core/activity/js/activity.js
```diff
@@ -122,8 +122,9 @@ define(['jquery', 'underscore', 'oae.core', 'activityadapter'], function($, _, o
             infinityScroll = $('.oae-list', $rootel).infiniteScroll(url, {
                 'limit': 10
             }, '#activity-items-template', {
-                'postProcessor': processActivities,
-                'emptyListProcessor': handleEmptyResultList
+                'ariaLive': 'off',  // Notifications will alert screen readers
```
nicolaasmatthijs: Not all new activities will cause a notification to be shown. We only show the notifications for new "notifications". That means a screenreader would currently not see any other new activities.

shared/oae/js/jquery-plugins/jquery.infinitescroll.js
```diff
@@ -34,6 +34,7 @@ define(['jquery', 'underscore', 'oae.api.util', 'oae.api.i18n', 'oae.api.l10n'],
      * @param  {Object}                            [options]                       Optional object containing additional configuraton options.
      * @param  {String|Element}                    [options.scrollContainer]       jQuery element or selector for that jQuery element that identifies the container on which the scrollposition should be watched to check when we are close enough to the bottom to load a new set of results. If this is not provided, the document body will be used.
      * @param  {String|Function}                   [options.initialContent]        HTML string that should be prepended to the list upon initialization. If a function is provided, the function will be called with no parameters and should return the HTML string to prepend.
+     * @param  {String         }                   [options.ariaLive]              `aria-live` attribute for container. Defaults to "assertive"
```
nicolaasmatthijs: Spacing issue

node_modules/oae-core/activity/js/activity.js
nicolaasmatthijs: Maybe we should set `'ariaLive': 'assertive'` to the element containing the title of the activity that was added?

shared/oae/js/jquery-plugins/jquery.infinitescroll.js
```diff
@@ -49,7 +50,9 @@ define(['jquery', 'underscore', 'oae.api.util', 'oae.api.i18n', 'oae.api.l10n'],
 
         // Default values
         parameters = parameters || {};
-        options = options || {};
+        options = $.extend({
```
nicolaasmatthijs: To be consistent, we should do:

```
options = options || {};

// Number of items to load per call to the server
parameters.limit = parameters.limit || 10;
// Aria live attribute to apply to the infinite scroll container
options.ariaLive = options.ariaLive || 'assertive';
...
```

nicolaasmatthijs:

node_modules/oae-core/activity/js/activity.js
simong: It's a bit tricky to determine whether an activity that gets placed in the activity stream gets pushed out as a notification as well in the UI though.
IIRC, the `activityId` between an activity and a notification will be different. So you can't compare on that.

Notifications are pushed to users on aggregation, which means you'd have to aggregate incoming activities with older activities (which may or may not be available to the UI) and then wait up to 5 seconds until a notification rolls in. Other than the delay implication, this would probably work for most use-cases, but would still be a pain as you'd then have to copy over all the backend aggregation rules.

Not reading a new activity (that has no notification) doesn't seem like such a big problem to me. We've already deemed it "not worthy" of being sent to the notification stream. Maybe it's ok to let the user discover it in their activity stream on a hard page refresh?

sathomas: Could it work to add a bit of code in the server? If I understand correctly (which may be doubtful :), the server knows whether any particular event will be pushed to the activity stream and/or the notifications stream. For those events that are going to be pushed to both streams, the server could add a flag to the event in the activity stream that, in effect, says "this event is also going to be pushed to the notifications stream."

The UI could then check for that flag on all events in the activity stream. If the flag is present, then it can suppress screen readers using `aria-live="off"` for that event.

I haven't thought through the implications of this approach w.r.t. aggregation, so it may not really work. But at first glance it seems feasible. (Of course, I've never looked at the server code, so I'm making a big assumption that it would be easy for the server to set such a flag.)

sathomas: Also, FWIW, if we do want to clue screen readers on some events in the activity stream (e.g. those that aren't notifications), I think `aria-live="polite"` is probably better than `aria-live="assertive"`. As @simong notes, we've already made the determination that those aren't critical events because we've omitted them from notifications.

sathomas:
Note that I've updated the PR with fixes for everything except the "not all activities are notifications" problem. If we want to follow the suggestion of @simong and simply not read any updates in the activity stream (at least as an interim fix?), then the PR could be reviewed now.

sathomas:
Need to fix content and discussion comments as well

nicolaasmatthijs:
@sathomas : Is any additional work required here?

sathomas:
@nicolaasmatthijs This can probably be reviewed and merged, since it's a definite improvement. It's conservative with respect to live regions, though. Basically, it disables screen reader live updates for the entire activity stream and for comments and discussion updates. The assumption is that any critical updates will be conveyed as push notifications, and the screen reader will pick up those notifications.

If we want to get more granular and only disable screen reader live updates (in the activity stream, comments, and discussions) for events that are definitely going to be push notifications as well, then I think we'll have to do some work on the server.

I'll go ahead and mark this PR as ready for @nicolaasmatthijs review.

nicolaasmatthijs:
@sathomas : Before I go ahead and merge this, would you be able to get in touch with Jared, explain him the rationale and check whether this would be acceptable in terms of a certification. I just want to make sure that we don't make the problem worse by not reading it at all.

sathomas:
I'll post an update when I hear back from WebAIM.

sathomas:
My query:

> It’s actually the case that new items in the users activity stream are not given the role of “alert”; however, the entire activity stream is labeled as an aria-live region. We’re assuming that the aria-live attribute was causing your screen reader to read the new activity stream content. Banner notifications do indeed have a role of “alert”.
>
> Our current approach to fixing this problem is exactly as you suggest. We’ve disabled the aria-live attribute for the activity streams content, and we’re relying exclusively on banner notifications with a role of “alert” to let screen readers know about the new activities. The quandary we face, though, is that some updates to the activity stream do not result in notifications at all. Those updates are ones we deem minor and not worthy of a notification. If we continue on the current course, those updates will be invisible to screen readers unless the user explicitly refreshes the page or discovers them through normal page navigation.
> 
> Do you have any opinions or guidance on this approach? We think that since we’ve already determined that these updates aren’t critical (since we’re not signaling them as banner notifications), that’s it’s best not to interrupt the screen reader user. But that does mean that the page might undergo (admittedly minor) updates without any notice to screen readers. Does that sound like a reasonably trade-off, or are we being too presumptuous in assuming that users wouldn’t want to be interrupted?

The WebAIM response:

> Very good questions. I think what you've described works very well. The intrusiveness of the frequent and lengthy activity stream updates probably outweigh the disadvantage of the user not being informed at all of new updates.
> 
> aria-live="polite" was originally intended to address these types of updates by informing the user that an update has occurred (perhaps via a beep or some other indicator), but not reading it immediately. Unfortunately there wasn't agreement on how to inform the user, so it was redefined to read the updates after the user pauses current activity.
> 
> So I think allowing the user to discover these less-important updates on their own works fine. Perhaps having an off-screen indicator before the activity stream of "x new updates below" or something like that could provide an indicator that updates have occurred without the user having to discover them.
> 
> Jared

PR #3911, opened 2015-01-13
(A11y severity 1) Investigate accessibility of preview documents

sathomas:
(Given the constraints of pdf2htmlEX this may not be fixable.)

The documents within the preview don’t read properly. Some sentences are read while entire sections are either skipped or no information is presented.

Also need to ensure any work here doesn't conflict with annotations.

### DIFF
--- a/node_modules/oae-core/documentpreview/documentpreview.html
+++ b/node_modules/oae-core/documentpreview/documentpreview.html
@@ -52,10 +52,13 @@
 
 <!-- TEMPLATES -->
 <div id="documentpreview-content-page-template"><!--
-    <div class="documentpreview-content-page ${cssScopeClass}" data-page-number="${pageNumber}">
-        <div class="documentpreview-content-page-loading text-center">
-            <i class="fa fa-spinner fa-spin"><span class="sr-only">__MSG__LOADING__</span></i>
+    <div class="documentpreview-content-page" data-page-number="${pageNumber}" >
+        <div class="documentpreview-content-html-page ${cssScopeClass}" aria-hidden="true">
+            <div class="documentpreview-content-page-loading text-center">
+                <i class="fa fa-spinner fa-spin"><span class="sr-only">__MSG__LOADING__</span></i>
+            </div>
         </div>
+        <div class="documentpreview-content-text-page sr-only" data-page-number="${pageNumber}"></div>
     </div>
 --></div>
 
--- a/node_modules/oae-core/documentpreview/js/documentpreview.js
+++ b/node_modules/oae-core/documentpreview/js/documentpreview.js
@@ -108,14 +108,14 @@ define(['jquery', 'underscore', 'oae.core', 'lazyload'], function($, _, oae) {
             page.height = page.$el.height();
             page.marginTop = CONTENT_PAGE_SPACING;
 
-            // Request the page content
+            // Request the page content HTML version
             $.ajax({
                 'url': constructDocumentPreviewURL('page.' + page.pageNumber + '.html'),
                 'crossDomain': true,
                 'dataType': 'text',
                 'success': function(response) {
                     // Replace the loading indicator with the page content
-                    page.$el.html(response);
+                    $('.documentpreview-content-html-page', page.$el).html(response);
 
                     // Update the height now that we have real content
                     page.height = page.$el.height();
@@ -162,6 +162,17 @@ define(['jquery', 'underscore', 'oae.core', 'lazyload'], function($, _, oae) {
 
                     // For all other error responses, we pass the error back up the call stack
                     return callback({'code': jqXHR.status, 'msg': jqXHR.responseText});
+                }
+            });
+
+            // Request the page content text version
+            $.ajax({
+                'url': constructDocumentPreviewURL('page.' + page.pageNumber + '.txt'),
+                'crossDomain': true,
+                'dataType': 'text',
+                'success': function(response) {
+                    // Add the text version of the content
+                    $('.documentpreview-content-text-page', page.$el).text(response);
                 }
             });
         };


sathomas:
This is a lot simpler than I was originally thinking, but it seems to work pretty well. There is an edge case that's not handled—if the signature expires after the request for a page HTML but before the request for a page text, then the text version won't get loaded. The chance of that happening seems vanishingly small, though, so I'm not sure if it's worth the complication to handle it. To @nicolaasmatthijs

node_modules/oae-core/documentpreview/documentpreview.html
```diff
@@ -52,10 +52,13 @@
 
 <!-- TEMPLATES -->
 <div id="documentpreview-content-page-template"><!--
-    <div class="documentpreview-content-page ${cssScopeClass}" data-page-number="${pageNumber}">
-        <div class="documentpreview-content-page-loading text-center">
-            <i class="fa fa-spinner fa-spin"><span class="sr-only">__MSG__LOADING__</span></i>
+    <div class="documentpreview-content-page" data-page-number="${pageNumber}" >
+        <div class="documentpreview-content-html-page ${cssScopeClass}" aria-hidden="true">
+            <div class="documentpreview-content-page-loading text-center">
+                <i class="fa fa-spinner fa-spin"><span class="sr-only">__MSG__LOADING__</span></i>
```
nicolaasmatthijs: I guess this means that the loading text will no longer be read as it is in an `aria-hidden` block? Would it be worth pulling the loading icon out of the `documentpreview-content-html-page` block?

node_modules/oae-core/documentpreview/js/documentpreview.js
```diff
@@ -164,6 +164,17 @@ define(['jquery', 'underscore', 'oae.core', 'lazyload'], function($, _, oae) {
                     return callback({'code': jqXHR.status, 'msg': jqXHR.responseText});
                 }
             });
+
+            // Request the page content text version
+            $.ajax({
+                'url': constructDocumentPreviewURL('page.' + page.pageNumber + '.txt'),
+                'crossDomain': true,
```
nicolaasmatthijs: I wonder why both of these require a `crossDomain: true` flag ?

node_modules/oae-core/documentpreview/js/documentpreview.js
```diff
@@ -164,6 +164,17 @@ define(['jquery', 'underscore', 'oae.core', 'lazyload'], function($, _, oae) {
                     return callback({'code': jqXHR.status, 'msg': jqXHR.responseText});
                 }
             });
+
+            // Request the page content text version
```
nicolaasmatthijs: Can you add here that this is for accessibility purposes?

node_modules/oae-core/documentpreview/js/documentpreview.js
sathomas: Me too. I'll dig through the commit history to see when it was added.

node_modules/oae-core/documentpreview/js/documentpreview.js
```diff
@@ -108,14 +108,14 @@ define(['jquery', 'underscore', 'oae.core', 'lazyload'], function($, _, oae) {
             page.height = page.$el.height();
             page.marginTop = CONTENT_PAGE_SPACING;
 
-            // Request the page content
+            // Request the page content HTML version
             $.ajax({
```
nicolaasmatthijs: Thinking about this, my preference would be to create a new endpoint that returns both the html and txt content as properties

```
html: "...",
text: "..."
```

If this is controlled by the signatures as well, we wouldn't have to worry about one of them expiring.

Would you be able to create a ticket for this and assign to @simong ?

nicolaasmatthijs:

node_modules/oae-core/documentpreview/js/documentpreview.js
simong: Any reason why they have to be combined? This wouldn't work really well with a storage backend such as S3 as the backend servers would have to act as a proxy and would suck up quite a bit of bandwidth/IO.

nicolaasmatthijs: Having them combined would be a lot cleaner from a UI point of view and wouldn't require as many individual requests. Having said that, I see your point about storage backends, so we should probably go for the separate requests approach here.

node_modules/oae-core/documentpreview/js/documentpreview.js
```diff
@@ -164,6 +164,17 @@ define(['jquery', 'underscore', 'oae.core', 'lazyload'], function($, _, oae) {
                     return callback({'code': jqXHR.status, 'msg': jqXHR.responseText});
                 }
             });
+
+            // Request the page content text version
+            $.ajax({
```
nicolaasmatthijs: It would probably be better to initiate the call to load the text version in the success block of the html version. That will ensure that our signature is always up to date. In order to do this cleanly, we should probably split up the `loadPage` function into a `renderPage` function that does the preparatory work, a `loadHTMLPage` function that loads the HTML page and a `loadTextPage` function that loads the text version.

node_modules/oae-core/documentpreview/js/documentpreview.js
sathomas: @nicolaasmatthijs:  Here's the [commit](https://github.com/oaeproject/3akai-ux/commit/db763f97770fd3279a9c51d9a469a14f99c037f9#diff-c9a9baad3b5d7cace1b766de4d0e2f47) ("Removed text preview and added crossDomain flag for Markdown preview"). I'm still not sure why this is needed, but maybe this will jog your memory?

sathomas:
Refactored the page loading code. This version handles the race condition of signature expiry in between HTML and text requests. To @nicolaasmatthijs for review

node_modules/oae-core/documentpreview/documentpreview.html
```diff
@@ -52,10 +52,12 @@
 
 <!-- TEMPLATES -->
 <div id="documentpreview-content-page-template"><!--
-    <div class="documentpreview-content-page ${cssScopeClass}" data-page-number="${pageNumber}">
+    <div class="documentpreview-content-page" data-page-number="${pageNumber}" >
```
nicolaasmatthijs: Shouldn't need a space at the end

node_modules/oae-core/documentpreview/documentpreview.html
```diff
@@ -52,10 +52,12 @@
...
         <div class="documentpreview-content-page-loading text-center">
             <i class="fa fa-spinner fa-spin"><span class="sr-only">__MSG__LOADING__</span></i>
         </div>
+        <div class="documentpreview-content-html-page ${cssScopeClass}" aria-hidden="true"></div>
+        <div class="documentpreview-content-text-page sr-only" data-page-number="${pageNumber}"></div>
```
nicolaasmatthijs: Do we need 2 separate elements with `data-page-number` ?

node_modules/oae-core/documentpreview/js/documentpreview.js
```diff
@@ -84,86 +171,85 @@ define(['jquery', 'underscore', 'oae.core', 'lazyload'], function($, _, oae) {
...
 
-            // Don't reload the page if it has already been loaded
-            if (page.$el) {
+            // Don't reload the page if all of its components have already been loaded
+            if (page.$el && page.$html && page.$text) {
                 return callback();
             }
+
+            // Variables to track progress of HTML and text content loads
+            var htmlDeferred = new $.Deferred();
```
nicolaasmatthijs: Whilst I can see why deferreds are being used here, I don't think we should start to mix them in in ad-hoc way and I'd prefer to take them back out of this PR

node_modules/oae-core/documentpreview/js/documentpreview.js
```diff
@@ -71,6 +71,93 @@ define(['jquery', 'underscore', 'oae.core', 'lazyload'], function($, _, oae) {
...
+            page.marginTop = CONTENT_PAGE_SPACING;
+        };
+
+        /**
+         * Render the HTML content for a page
+         *
+         * @param  {Object}         page                Page object representing the page that will be loaded
+         * @param  {String}         html                HTML content for the page
+         */
+        var renderPageHtml = function(page, html) {
```
nicolaasmatthijs: I would fetch the HTML content as part of this function, and pass in a callback to call when the HTML page has been loaded and rendered

node_modules/oae-core/documentpreview/js/documentpreview.js
```diff
@@ -71,6 +71,93 @@ define(['jquery', 'underscore', 'oae.core', 'lazyload'], function($, _, oae) {
...
+                'cssScopeClass': widgetData.previews.cssScopeClass,
+                'pageNumber': page.pageNumber
+            }));
+
+            // Cache a reference to the page element
+            page.$el = $('.documentpreview-content-page[data-page-number="' + page.pageNumber + '"]', $rootel);
+
+            // Cache the height and calculated top margin of the
+            // element as they are needed for zoom calculations
+            page.height = page.$el.height();
```
nicolaasmatthijs: Why do we cache the height before the page content is loaded?

node_modules/oae-core/documentpreview/js/documentpreview.js
```diff
@@ -71,6 +71,93 @@ define(['jquery', 'underscore', 'oae.core', 'lazyload'], function($, _, oae) {
...
+            zoomPages(page.pageNumber);
+        };
+
+        /**
+         * Render the text content for a page (for screen readers)
+         *
+         * @param  {Object}         page                Page object representing the page that will be loaded
+         * @param  {String}         text                Text content for the page
+         */
+        var renderPageText = function(page, text) {
```
nicolaasmatthijs: I would fetch the text content as part of this function, and pass in a callback to call when the text page has been loaded and rendered

node_modules/oae-core/documentpreview/js/documentpreview.js
```diff
@@ -71,6 +71,93 @@ define(['jquery', 'underscore', 'oae.core', 'lazyload'], function($, _, oae) {
...
+            // Add the text content
+            page.$text = $('.documentpreview-content-text-page', page.$el).text(text);
+        };
+
+        /**
+         * Handle page loading errors
```
nicolaasmatthijs: Should explain why we need to handle this

node_modules/oae-core/documentpreview/js/documentpreview.js
nicolaasmatthijs: That means we should roughly have something like this:

```
prepareForPageLoad();
loadHTMLPage(function() {
  loadTextPage();
});
```

node_modules/oae-core/documentpreview/js/documentpreview.js
```diff
@@ -84,86 +171,85 @@ define(['jquery', 'underscore', 'oae.core', 'lazyload'], function($, _, oae) {
...
+                    'dataType': 'text',
+                    'success': function(response) {
+                        // Render the HTML content for the page
+                        renderPageHtml(page, response);
+
+                        // Indicate that the page has finished loading, so more pages can be loaded
+                        isLoadingPage = false;
+
+                        // Check if a new page should be loaded in case we're still close to the bottom of the container
+                        if (!isLoadingPages) {
```
nicolaasmatthijs: This should only be called when the text version has also been fetched

node_modules/oae-core/documentpreview/js/documentpreview.js
```diff
@@ -84,86 +171,85 @@ define(['jquery', 'underscore', 'oae.core', 'lazyload'], function($, _, oae) {
...
+                $.ajax({
+                    'url': constructDocumentPreviewURL('page.' + page.pageNumber + '.html'),
+                    'crossDomain': true,
+                    'dataType': 'text',
+                    'success': function(response) {
+                        // Render the HTML content for the page
+                        renderPageHtml(page, response);
+
+                        // Indicate that the page has finished loading, so more pages can be loaded
+                        isLoadingPage = false;
```
nicolaasmatthijs: This should only be set to false when the text version has also been retrieved

nicolaasmatthijs:

node_modules/oae-core/documentpreview/js/documentpreview.js
sathomas: The problem with that approach is that it serializes the two AJAX requests, thus adding an additional round trip delay to page rendering. It seems like that might be a pretty high cost (especially for lengthy documents) as the document preview content page is pretty slow to load and render already.

Accessing `oae.gatech.edu` from Atlanta shows about 400-600ms of latency per request, so this would add about half a second of delay for each page in the document.

node_modules/oae-core/documentpreview/js/documentpreview.js
sathomas: In case the user scrolls the document while the page content is being retrieved. In that case the height is just the height of the loading message, but we still need to keep track of it.

node_modules/oae-core/documentpreview/documentpreview.html
```diff
@@ -52,9 +52,17 @@
...
 <div id="documentpreview-content-page-template"><!--
-    <div class="documentpreview-content-page ${cssScopeClass}" data-page-number="${pageNumber}">
-        <div class="documentpreview-content-page-loading text-center">
-            <i class="fa fa-spinner fa-spin"><span class="sr-only">__MSG__LOADING__</span></i>
+    <div class="documentpreview-content-page" data-page-number="${pageNumber}">
+        <div class="documentpreview-content-text-page sr-only"">
+            <label for="documentpreview-content-text-page-${pageNumber}">
+                __MSG__PAGE_X_OF_Y_PREFIX__ ${pageNumber} __MSG__PAGE_X_OF_Y_MIDDLE__ ${totalPages} __MSG__PAGE_X_OF_Y_SUFFIX__
+            </label>
+            <textarea readonly id="documentpreview-content-text-page-${pageNumber}">__MSG__LOADING__</textarea>
```
nicolaasmatthijs: Is it a semantic issue for screenreader users to use a form field for something that is basically a read-only display?

node_modules/oae-core/documentpreview/documentpreview.html
```diff
@@ -52,9 +52,17 @@
...
+        </div>
+        <div class="documentpreview-content-html-page ${cssScopeClass}" aria-hidden="true">
+            <div class="documentpreview-content-page-loading text-center">
```
nicolaasmatthijs: Can you make sure that this is prefixed with `content-html-page` as well?

node_modules/oae-core/documentpreview/js/documentpreview.js
```diff
@@ -71,6 +71,95 @@ define(['jquery', 'underscore', 'oae.core', 'lazyload'], function($, _, oae) {
...
+            // If we could not load the page due to a 401 (access denied) and it's the first time we tried loading this
+            // particular page, we refresh the signature and try again
+            if (jqXHR.status === 401 && !_isRetryAttempt) {
+                refreshSignature(function(err) {
+                    if (err) {
+                        callback(err);
+                    }
+
+                    // Try loading the page again
+                    loadPage(page, callback, true);
```
nicolaasmatthijs: I think there's currently a chance that we'd load the same page twice as the signature can be refreshed twice. The code would detect that the signature is out of date, and depending on the timing of the requests, could call handlePageLoadError twice.

node_modules/oae-core/documentpreview/js/documentpreview.js
nicolaasmatthijs: That makes sense. Lets go with the following pattern then:

```
var loadPage = function(page, callback, etc.) {
   // preload page
   // load html page
   loadPageHTML(page, function) {
    callback();
    loadPageText(page);
  }
}
```

That ensures that the HTML pages are loaded as fast as possible, and brings in the text version asynchronously. It also means that we don't really have to take the signature expiry scenario into account for the text loading function (as the html loading always just precedes it).

nicolaasmatthijs:

node_modules/oae-core/documentpreview/documentpreview.html
sathomas: I haven't found any suggestions that this is a problem, and it works perfectly with VoiceOver. Using a `<textarea>` puts the page into the tab focus group, so users can tab from one page to the next. (It works for sighted users as well; tabbing scrolls the page into view.)

sathomas:
refactored as callbacks; over to @nicolaasmatthijs

node_modules/oae-core/documentpreview/documentpreview.html
nicolaasmatthijs: Still needs to be taken care of

node_modules/oae-core/documentpreview/js/documentpreview.js
```diff
@@ -71,51 +71,65 @@ define(['jquery', 'underscore', 'oae.core', 'lazyload'], function($, _, oae) {
...
 
             // Cache a reference to the page element
             page.$el = $('.documentpreview-content-page[data-page-number="' + page.pageNumber + '"]', $rootel);
 
             // Cache the height and calculated top margin of the
-            // element as they are needed for zoom calculations
+            // element as they are needed for zoom calculations.
+            // Initially the height is just the height of the loading
+            // message, but that's still critical for scaling and
+            // scrolling correctly.
```
nicolaasmatthijs: Remove dot at end of `//` comment (block)

node_modules/oae-core/documentpreview/js/documentpreview.js
```diff
@@ -71,51 +71,65 @@ define(['jquery', 'underscore', 'oae.core', 'lazyload'], function($, _, oae) {
...
+            $.ajax({
+                'url': constructDocumentPreviewURL('page.' + page.pageNumber + '.txt'),
+                'crossDomain': true,
+                'dataType': 'text',
+                'success': function(response) {
+                    page.$text = $('.documentpreview-content-text-page', page.$el)
+                    $('textarea', page.$text).text(response);
+                }
+                // Don't handle error responses. The only recoverable response is a
+                // signature expiry and loading the HTML content will catch that.
```
nicolaasmatthijs: Same here

node_modules/oae-core/documentpreview/js/documentpreview.js
```diff
@@ -71,51 +71,65 @@ define(['jquery', 'underscore', 'oae.core', 'lazyload'], function($, _, oae) {
...
+            });
+        };
+
+        /**
+         * Load and render the HTML content for a page
+         *
+         * @param  {Object}         page                Page object representing the page that will be loaded
+         * @param  {Function}       [callback]          Standard callback function
```
nicolaasmatthijs: The callback is always executed, so it's not optional

node_modules/oae-core/documentpreview/js/documentpreview.js
```diff
@@ -131,39 +145,99 @@ define(['jquery', 'underscore', 'oae.core', 'lazyload'], function($, _, oae) {
...
+
+            // For all other error responses, we pass the error back up the call stack
+            callback({'code': jqXHR.status, 'msg': jqXHR.responseText});
+        };
+
+        /**
+         * Load a document page and append it to the document viewer
+         *
+         * @param  {Object}         page                Page object representing the page that needs to be loaded and rendered
+         * @param  {Function}       [callback]          Standard callback function
```
nicolaasmatthijs: The callback is always executed, so it can't be optional

node_modules/oae-core/documentpreview/js/documentpreview.js
```diff
@@ -131,39 +145,99 @@ define(['jquery', 'underscore', 'oae.core', 'lazyload'], function($, _, oae) {
...
+         * @param  {Boolean}        [_isRetryAttempt]   `true` if we attempted to load this page previously
+         */
+        var loadPage = function(page, callback, _isRetryAttempt) {
+            // Set a default callback function in case no callback function has been provided
+            callback = callback || function() {};
+            // Indicates whether or not we attempted to load this page previously
+            _isRetryAttempt = _isRetryAttempt || false;
+
+            // Don't reload the page if all of its components have already been loaded
+            if (page.$el && page.$html && page.$text) {
```
nicolaasmatthijs: What's the scenario in which this would be triggered?

node_modules/oae-core/documentpreview/js/documentpreview.js
```diff
@@ -131,39 +145,99 @@ define(['jquery', 'underscore', 'oae.core', 'lazyload'], function($, _, oae) {
...
+
+            // Keep track of the last page in the document that was loaded
+            lastPageLoaded = page.pageNumber;
+
+            // Set up the HTML markup for the page
+            prepareForPageLoading(page);
+
+            // Kick off the request for the page's text content (for screen readers).
+            // Load the text first so we can rely on the HTML load to catch signature
+            // expiry and initiate a recovery.
```
nicolaasmatthijs: Same here ?

node_modules/oae-core/documentpreview/js/documentpreview.js
```diff
@@ -131,39 +145,99 @@ define(['jquery', 'underscore', 'oae.core', 'lazyload'], function($, _, oae) {
...
+            if (!page.$text) {
```
nicolaasmatthijs: I think we want the HTML page to be loaded first, as that will refresh the signature.

```
loadPageHtml(page, function(jqXHR) {
  ...
  loadPageText(page);
  ...
  callback();
```

node_modules/oae-core/documentpreview/js/documentpreview.js
```diff
@@ -131,39 +145,99 @@ define(['jquery', 'underscore', 'oae.core', 'lazyload'], function($, _, oae) {
...
+                    // Returned object indicates an error
+                    if (jqXHR) {
+                        // Handle the specific load error
+                        handlePageLoadError(page, callback, _isRetryAttempt, jqXHR);
+                    } else {
+                        // Indicate that the page has finished loading, so more pages can be loaded
+                        isLoadingPage = false;
+
+                        // Check if a new page should be loaded in case we're still close to the bottom of the container
+                        if (!isLoadingPages) {
```
nicolaasmatthijs: Where does `isLoadingPages` come from? Is it a typo from `isLoadingPage`? If so, `isLoadingPage` is always going to be false because of the line above.

nicolaasmatthijs:
Left some more review. Back to @sathomas for follow-up.

node_modules/oae-core/documentpreview/js/documentpreview.js
sathomas: If I understand the "signature" concept (which is a tenuous assumption), that would still leave open the possibility of a race condition.

1. User pauses a while to read the currently displayed preview page.
2. Just **before** the signature expires, the user advances to a new page.
3. The HTML ajax request just makes it "under the gun" before the signature expires; the request succeeds.
4. The text ajax request which follows just misses the expiration time and fails.

Honestly, I think `$.Deferred` (or vanilla Promises with a polyfill) are the cleanest way to handle these scenarios, but if we don't want to use them, then we need to place the error recovery (i.e. signature refresh) in the last ajax request instead of the first. That means either (a) making the HTML request after the text request (as in the above), or (b) moving all the error recovery code to the text request. More generically, the scenario becomes.

1. User pauses a while to read the currently displayed preview page.
2. Just **before** the signature expires, the user advances to a new page.
3. The first ajax request just makes it "under the gun" before the signature expires; the request succeeds.
4. The second ajax request which follows just misses the expiration time and fails; signature refresh kicks in and the retry of the request succeeds.

This logic also handles the more typical case of the signature expiration before either request, although it's not quite as easy to see.

1. User pauses a while to read the currently displayed preview page.
2. After the signature expires, the user advances to a new page.
3. The first ajax request fails because of an expired signature and, since no error recovery is invoked, has no effect. (i.e. the failure is silent)
4. The second ajax request also fails; for this request, however, signature refresh kicks in both request are retried and succeed.

node_modules/oae-core/documentpreview/js/documentpreview.js
sathomas: User enters page number, e.g. 50, but, before that page is loaded, gets impatient and enters a new page number, e.g. 60. Some of the pages between page 1 and page 60 might have been loaded already because of the "aborted" request to advance to page 50.

node_modules/oae-core/documentpreview/js/documentpreview.js
sathomas: `isLoadingPage` ➞ an ajax request (or two, in the case of both HTML and text) is in progress
`isLoadingPages` ➞ multiple pages need to be loaded to "get to" the page the user has requested

node_modules/oae-core/documentpreview/js/documentpreview.js
sathomas: Adding to the note above, I don't think it's practical to load the HTML first and move the error handling/recovery to the text request. Because the error handling also deals with the callback, that could result in the callback being called before the HTML is loaded. (Even though the HTML is requested first, the text request could finish ahead of it.) In that scenario, we know that the HTML will eventually be loaded (since the text request has succeeded), but it's still possible that the callback will run before that happens, If we're loading multiple pages, then the callback code go ahead to the next page load. When that page load starts, it calculates the page geometry to handle scrolling and zooming. In this example, though, the page geometry calculation will be wrong because the preceding HTML content hasn't yet been loaded.

sathomas:
I think everything's addressed in the follow-up, but my memory is hazy so I may be missing something. Back to @nicolaasmatthijs for review.

BTW the Travis CI `jshint` warning makes absolutely no sense:

```
shared/oae/api/oae.api.group.js
    310 |    }
              ^ Missing semicolon.
```

That file only has 284 lines total in it. Running `grunt jshint` locally shows no errors.

PR #3964, opened 2015-01-13
Display issue for new comments

sathomas:
On mobile devices, we give second and third level replies less margin-left compared to larger screens:

screen shot 2015-01-12 at 22 20 08

However, when entering the comment, these rules don't appear to be respected. The comment box should be aligned with where the comment will show up.

screen shot 2015-01-12 at 22 19 58

There is also a secondary issue when replying to 3rd level comments. A reply to a 3rd level comment will show as a 3rd level comment as well. Therefore, that reply box should be shown in the same place as where the comment will end up. However, the reply box is currently indented to level 4.

screen shot 2015-01-12 at 22 23 40

### DIFF
--- a/node_modules/oae-core/comments/css/comments.css
+++ b/node_modules/oae-core/comments/css/comments.css
@@ -40,6 +40,15 @@
 
 /* iPhone resolution */
 @media (max-width: 480px) {
+    .comments-widget #comments-container > .media > .media-body {
+        overflow: visible;
+    }
+
+    .comments-widget .comments-level-0 .comments-reply-container,
+    .comments-widget .comments-level-1 .comments-reply-container {
+        margin-left: 35px;
+    }
+
     .comments-widget .comments-level-1 {
         margin-left: 35px;
     }


sathomas:
I think this matches the requirements. To @nicolaasmatthijs to check.

nicolaasmatthijs:
There are still some functional issues with this:

- The input box for 4th level replies is still indented, whilst the actual 4th level reply comment is not. 4th level replies (and beyond) should all be shown at level 3, including the input box. Not that this is working as expected on mobile devices, but not on larger devices.

- The new comment box is now cut off on the right side

screen shot 2015-01-14 at 07 18 46

node_modules/oae-core/comments/css/comments.css
```diff
@@ -40,6 +40,15 @@
 
 /* iPhone resolution */
 @media (max-width: 480px) {
+    .comments-widget #comments-container > .media > .media-body {
+        overflow: visible;
+    }
+
+    .comments-widget .comments-level-0 .comments-reply-container,
```
nicolaasmatthijs: If we need this, it should be combined with the rule below

nicolaasmatthijs:

mrvisser:
In the comment picture above (https://cloud.githubusercontent.com/assets/109850/5741000/2dbda6e8-9bbd-11e4-85d6-2188633395b0.png) the top padding above `Your Comment` seems excessive, making it look asymetric to me, not sure if that's intended.

node_modules/oae-core/comments/css/comments.css
sathomas: @nicolaasmatthijs:

> Not that this is working as expected on mobile devices, but not on larger devices.

Was the first word intended to be "Note" instead of "Not"? I ask because the only changes in this PR are in the mobile device media query, so if it doesn't work on larger devices, that's been a problem all along.

nicolaasmatthijs: @sathomas : Yes, that was supposed to say "Note". I'm just suggesting that we include a fix for larger devices in this PR as well.

sathomas: Got it. Just wanted to be sure I was looking for the right problem

sathomas:
That was a lot of CSS pain just because Bootstrap 3 added one `overflow: hidden`. Back to @nicolaasmatthijs to check

nicolaasmatthijs:
There is a remaining issue with this. When the `x time ago` indication is long enough and the actual comment is pushed down, the comment shows below the profile picture, whilst it should be showing to the right of it.

screen shot 2015-01-19 at 19 16 03

nicolaasmatthijs:

sathomas:
tweaked the margin setting; back to @nicolaasmatthijs

PR #3965, opened 2015-01-16
Skinning error in configurable landing pages

nicolaasmatthijs:
Skinning isn't being applied correctly to the lower left block. 

screen shot 2015-01-16 at 10 09 56 am

### DIFF
--- a/ui/js/index.js
+++ b/ui/js/index.js
@@ -45,7 +45,7 @@ require(['jquery','oae.core'], function($, oae) {
                 // after all Markdown content has been converted to HTML
                 $.each(blocks, function(blockIndex, block) {
                     if (block.titleColor) {
-                        $('.index-block-' + blockIndex).filter(':header').css('color', block.titleColor);
+                        $('.index-block-' + blockIndex).find(':header').css('color', block.titleColor);
                     }
                 });
 


nicolaasmatthijs:
Added a fix. Over to @sathomas for review.

sathomas:
I think there's still an issue with the icon colors. For Georgia Tech, I think the icon color should be `#eeb211` since that's the hover color and the color of the play control in the video. Or maybe it's supposed to be gray? (At the very least, it shouldn't be blue.) Back to @nicolaasmatthijs

nicolaasmatthijs:
That's not a bug. The title color for that block is configured to be blue, which means that the icon will be blue as well. Feel free to reconfigure the title color for that block to the yellow of the icons.

nicolaasmatthijs:
Back to @sathomas for review

sathomas:
Sorry if I'm missing something, but there is absolutely nothing in the current Georgia Tech skin that is configured as blue. Is there maybe a variable that isn't exposed to the admin skinning interface that should be?

voila_capture 2015-01-16_10-46-30_am

nicolaasmatthijs:
It's part of the landing page configuration under Modules --> OAE Tenant Module. A proper admin UI for this is scheduled for the next release.

sathomas:
Ah. Got it. Fixed the Georgia Tech page and merged the PR.

PR #3966, opened 2015-01-21
Top navigation buttons should have tooltip

sathomas:
All top navigation buttons should have a standard tooltip (e.g. Notifications, Sign out, etc.)

### DIFF
--- a/node_modules/oae-core/topnavigation/js/topnavigation.js
+++ b/node_modules/oae-core/topnavigation/js/topnavigation.js
@@ -209,6 +209,15 @@ define(['jquery', 'oae.core', 'activityadapter'], function($, oae, ActivityAdapt
         };
 
         /**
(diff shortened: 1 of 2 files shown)


sathomas:
I added a tooltip to the search icon as well. Not sure if that's desirable. Also, I positioned the tips below the icons so that the home tooltip doesn't obscure the notifications tooltip. That may not be ideal either. Ready for @nicolaasmatthijs to comment/review

nicolaasmatthijs:
@sathomas : There's no need to use the Bootstrap tooltips for this. Would you be able to change this to use the standard title attribute tooltips?

sathomas:
Home and Notifications already had `title` attributes. They're back now, along with added `title` attributes for Search and Sign out. To @nicolaasmatthijs

nicolaasmatthijs:
Looks good. Merged.

PR #3969, opened 2015-01-28
Adjust 'return' label on iOS

sathomas:
When using iOS, the label on the submit button in the iOS keyboard reads "return". Is it possible to modify this to make it say "Search" instead?

screen shot 2015-01-27 at 20 54 26

### DIFF
--- a/node_modules/oae-core/topnavigation/topnavigation.html
+++ b/node_modules/oae-core/topnavigation/topnavigation.html
@@ -13,7 +13,7 @@
         <!-- RIGHT MENU -->
         <div id="topnavigation-right" class="col-xs-5 text-right">
             <!-- SEARCH FORM -->
-            <form id="topnavigation-search-form" class="form-inline" role="search" aria-label="__MSG__TOPNAV_SEARCH_ARIA_LABEL__">
+            <form action="." id="topnavigation-search-form" class="form-inline" role="search" aria-label="__MSG__TOPNAV_SEARCH_ARIA_LABEL__">
                 <div class="form-group">
                     <label for="topnavigation-search-query" class="control-label sr-only">__MSG__SEARCH__</label>
                     <input type="text" id="topnavigation-search-query" name="topnavigation-search-query" class="search-query form-control pull-left" title="__MSG__SEARCH_FOR_CONTENT_FOLDERS_PEOPLE_GROUPS__"/>
--- a/ui/index.html
+++ b/ui/index.html
@@ -28,7 +28,7 @@
                 {macro renderSearchBlock()}
                     <form action="." class="index-search-form" role="search" aria-label="__MSG__SEARCH_FOR_CONTENT_FOLDERS_PEOPLE_GROUPS__">
                         <div class="form-group">
-                            <input type="text" class="form-control search-query form-control pull-left index-search-query" title="__MSG__SEARCH_FOR_CONTENT_FOLDERS_PEOPLE_GROUPS__" placeholder="__MSG__SEARCH_FOR_CONTENT_FOLDERS_PEOPLE_GROUPS__"/>
+                            <input type="text" class="form-control search-query form-control pull-left index-search-query" title="__MSG__SEARCH_FOR_CONTENT_FOLDERS_PEOPLE_GROUPS__" placeholder="__MSG__SEARCH_FOR_CONTENT_FOLDERS_PEOPLE_GROUPS__" name="index-search"/>
                             <button type="submit" class="btn btn-link pull-left index-search-icon">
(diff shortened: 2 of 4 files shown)


sathomas:
With iOS 8, the search form has to have an `action` attribute to trigger the special search keyboard. Added in this PR; no telling what new hoops Apple with add with iOS 9. To @nicolaasmatthijs for review

ui/search.html
```diff
@@ -34,7 +34,7 @@
                             <div class="oae-clip-fold-outer"></div>
                         </div>
                         <div class="clearfix oae-clip-content">
-                            <form id="search-form" class="form-search form-horizontal" role="search" aria-label="__MSG__SEARCH__">
+                            <form action="." id="search-form" class="form-search form-horizontal" role="search" aria-label="__MSG__SEARCH__">
```
nicolaasmatthijs: @sathomas : Can we do the same for the other search forms in the application (landing page, top navigation, etc.) ?

nicolaasmatthijs:

sathomas:
added the attribute to topnavigation and index page; there's also a search field in `custom/thefold.oaeproject.org/index.html` which I didn't change. Are we still using that page, or have the configurable landing pages superseded it? To @nicolaasmatthijs

node_modules/oae-core/topnavigation/topnavigation.html
```diff
@@ -13,7 +13,7 @@
         <!-- RIGHT MENU -->
         <div id="topnavigation-right" class="col-xs-5 text-right">
             <!-- SEARCH FORM -->
-            <form id="topnavigation-search-form" class="form-inline" role="search" aria-label="__MSG__TOPNAV_SEARCH_ARIA_LABEL__">
+            <form action="." id="topnavigation-search-form" class="form-inline" role="search" aria-label="__MSG__TOPNAV_SEARCH_ARIA_LABEL__">
```
nicolaasmatthijs: @sathomas : I believe there are more search forms that would need to be adjusted (e.g. library searches, admin UI, etc.).

nicolaasmatthijs:

node_modules/oae-core/topnavigation/topnavigation.html
sathomas: Ah, missed those being set up in the list macro.

sathomas:
added the hint to the list macro. back to @nicolaasmatthijs

ui/index.html
```diff
@@ -26,7 +26,7 @@
             <div id="index-page-container" class="row"><!-- --></div>
             <div id="index-page-template"><!--
                 {macro renderSearchBlock()}
-                    <form class="index-search-form" role="search" aria-label="__MSG__SEARCH_FOR_CONTENT_FOLDERS_PEOPLE_GROUPS__">
+                    <form action="." class="index-search-form" role="search" aria-label="__MSG__SEARCH_FOR_CONTENT_FOLDERS_PEOPLE_GROUPS__">
```
nicolaasmatthijs: When I try this one on the iOS simulator, it shows "Go" instead of "Search". It's working correctly for the others though. Any idea what's causing this?

screen shot 2015-01-29 at 06 52 06

nicolaasmatthijs:
Left a comment. Back to @sathomas for follow-up.

sathomas:
Turns out you also have to have "search" in the `name` or `id` attribute. It would be nice if Apple would document these requirements somewhere. But then again, if they did that they wouldn't be able to arbitrarily change them as easily :) @nicolaasmatthijs to see if this resolves it.

ui/index.html
```diff
@@ -26,9 +26,9 @@
...
                         <div class="form-group">
-                            <input type="text" class="form-control search-query form-control pull-left index-search-query" title="__MSG__SEARCH_FOR_CONTENT_FOLDERS_PEOPLE_GROUPS__" placeholder="__MSG__SEARCH_FOR_CONTENT_FOLDERS_PEOPLE_GROUPS__"/>
+                            <input type="text" class="form-control search-query form-control pull-left index-search-query" title="__MSG__SEARCH_FOR_CONTENT_FOLDERS_PEOPLE_GROUPS__" placeholder="__MSG__SEARCH_FOR_CONTENT_FOLDERS_PEOPLE_GROUPS__" name="index-search"/>
```
nicolaasmatthijs: @sathomas : It is technically possible to configure your landing page with multiple search boxes. It would probably be best to ensure that the name is unique.

nicolaasmatthijs:
Added one final comment. Back to @sathomas

sathomas:
okay, added unique `id` value to search field name. to @nicolaasmatthijs

nicolaasmatthijs:
Looks good. Merged.

PR #3913, opened 2015-01-28
(A11y severity 2) Hide collapsed menu options from screen readers

sathomas:
When the display of the site is reduced, a menu icon appears to allow users to expand or contract the menu options. Although the menu is collapsed by default, users still navigate through all of the hidden options such as "Upload" before they can even access the icon to expand and contract the menu. Not only does the navigation order need to be changed to allow for the icon to be accessed before the menu options, but the options should be inaccessible to keyboard users (i.e., hidden with CSS display:none) when the menu is collapsed. This ensures that sighted keyboard users won’t navigate through elements that they cannot see.

### DIFF
--- a/shared/oae/css/oae.components.css
+++ b/shared/oae/css/oae.components.css
@@ -936,6 +936,15 @@ div.oae-thumbnail i.fa {
 
 /* Left-hand navigation collapsed view */
 @media (max-width: 991px) {
+    /* hide from screen readers */
+    .oae-lhnavigation {
+        display: none;
+    }
+
+    .oae-lhnavigation.oae-lhnav-expanded {
+        display: block;
+    }
+
     .oae-page {
         margin-left: 0;
         padding-left: 25px;
--- a/node_modules/oae-core/lhnavigation/lhnavigation.html
+++ b/node_modules/oae-core/lhnavigation/lhnavigation.html
@@ -2,7 +2,7 @@
 <link rel="stylesheet" type="text/css" href="css/lhnavigation.css"/>
 
 <div id="lhnavigation-container" class="row">
-    <div class="oae-lhnavigation">
+    <div class="oae-lhnavigation" id="oae-lhnavigation" aria-hidden="true">
         <ul class="nav nav-list"><!-- --></ul>
     </div>
     <div class="oae-page"><!-- --></div>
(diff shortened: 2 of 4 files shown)


sathomas:
To elaborate a bit, the issue with our current structure is that the hamburger icon is located within the main page container while all the actions are in the lhnav container. To meet the accessibility requirements, the hamburger icon and the actions need to be adjacent to each other within the markup. I think the easiest way to do that would be to move the hamburger icon from the main page container into the lhnav container. If we do that, it would certainly be a lot easier if we change the responsive design slightly to keep the icon in that container visually as well as in the markup. To see the effect, here's the current expanded view:

oldexpanded

If we move the icon, the new expanded view would be:

expanded

(I actually prefer this layout, as it means users can toggle the lhnav menu on and off by tapping the same location on the screen. With the current approach, I bet I've hit the Upload action a hundred times when I meant to close the menu.)

As a possible, but not required, change, we could also have the menu icons (but not the text) "peeking" even when the menu is collapsed. Current layout in collapsed state:

oldcollapsed

Possible adjustment:

collapsed

This could be something we'd do on intermediate layouts (e.g. iPads/tablets) since the extra screen real estate lost to the icons would be minimal. On smartphones we could continue with the current approach of hiding the entire menu.

sathomas:
Additional note: even if we don't want to change the markup/design, it's still worthwhile to merge this PR. It fixes the problem of tabbing to hidden elements and it adds the appropriate aria details to the menu.

shared/oae/css/oae.components.css
```diff
@@ -936,6 +936,15 @@ div.oae-thumbnail i.fa {
 
 /* Left-hand navigation collapsed view */
 @media (max-width: 991px) {
+    /* hide from screen readers */
+    .oae-lhnavigation {
+        display: none;
```
nicolaasmatthijs: This changes the animation slightly in that you can see the left hand navigation disappear before the animation is completed. Can we modify this to add a `hide` class when the animation is completed and remove the `hide` class when we start opening the navigation?

shared/oae/css/oae.components.css
```diff
@@ -936,6 +936,15 @@ div.oae-thumbnail i.fa {
 
 /* Left-hand navigation collapsed view */
 @media (max-width: 991px) {
+    /* hide from screen readers */
```
nicolaasmatthijs: Capital case to start sentence

nicolaasmatthijs:
Finished review for the current fix. Back to @sathomas for follow-up.

shared/oae/css/oae.components.css
nicolaasmatthijs: Actually, scrap the `hide` class as we only want this to happen on <991px devices. However, the `display:none` should still only be applied after the animation has finished.

sathomas:
Followed up on review. Back to @nicolaasmatthijs

PR #3939, opened 2015-01-28
(A11y severity 3) Hide close buttons from screen readers

sathomas:
The "close" buttons (), although they are marked up with `aria-hidden="true"`, are still navigated to by screen readers with no content read for them. This is because standard navigable elements (such as `<button>`) cannot be hidden with `aria-hidden`. Address this by removing `aria=hidden="true"` or by replacing the `<button>` elements with elements that will not receive keyboard focus (e.g., `<div onclick…>`) and maintaining `aria-hidden`.


sathomas:
I think I caught them all. To @nicolaasmatthijs for review

nicolaasmatthijs:
Looks good. Merged.

PR #3951, opened 2015-01-28
(A11y severity 3) Add markup hierarchy to comment threads

sathomas:
While nested comments are apparent visually, there is no indication of these relationships available to screen readers.
 A nested relationship could be conveyed through the use of either unordered lists or by heading structure. Unordered lists may be the most appropriate in this situation since the first comment begins with a fourth-level heading which doesn’t leave much room for sub-headings. Using unordered lists would provide the nesting concept by ensuring that each sub-comment is a list item for each parent comment. No changes to the visual presentation would be necessary (styles can be used to suppress the list bullets, etc.). Present the nested relationship information to users in a way that conveys the same information available visually.

### DIFF
--- a/node_modules/oae-core/comments/comments.html
+++ b/node_modules/oae-core/comments/comments.html
@@ -3,7 +3,7 @@
 
 <!-- CONTENT -->
 <div class="comments-widget">
-    <ul id="comments-container" class="media-list"><!-- --></ul>
+    <ul id="comments-container" class="media-list" role="list"><!-- --></ul>
 </div>
 
 <div id="comments-new-comment-template"><!--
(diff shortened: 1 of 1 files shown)


sathomas:
even better, using `role="list item"`. To @nicolaasmatthijs for review

node_modules/oae-core/comments/comments.html
```diff
@@ -3,7 +3,7 @@
 
 <!-- CONTENT -->
 <div class="comments-widget">
-    <ul id="comments-container" class="media-list"><!-- --></ul>
+    <ul id="comments-container" class="media-list" role="list"><!-- --></ul>
```
nicolaasmatthijs: The new comment box will be a part of the same list. Does that need to be labelled in any way?

nicolaasmatthijs:

node_modules/oae-core/comments/comments.html
sathomas: I thought about this for awhile and concluded that it's best not to add the aria/role markup to the new comment grouping. The point of the aria/role markup is to expose the hierarchy of the comments, and a new comment form (which isn't even a comment yet) doesn't really fit in that hierarchy. I then checked the W3C specs to make sure that there was no _requirement_ to mark all `<li>` as `role="listitem"` and there isn't. The flow in screenreaders seems okay as well. First it announces a list, then it reads the new comment box as a list item but without a level. Then it reads the existing comments with their associated levels.

sathomas:
I think this is okay as implemented, back to @nicolaasmatthijs for thoughts

nicolaasmatthijs:
Thanks for the clarification. Merged.

PR #3970, opened 2015-01-29
Add editor role to the UI

stuartf:
- When managing the members of a collaborative document, the "Can edit" role should be available as one of the roles
- When sharing a collaborative document as a manager, the editor role should be available as one of the roles in the dropdown
- When sharing a collaborative document as an editor, the editor role should be available as one of the roles in the dropdown
- When creating a new collaborative document, all groups and users the item is shared with should become editors

### DIFF
--- a/ui/js/content.js
+++ b/ui/js/content.js
@@ -307,7 +307,7 @@ require(['jquery', 'underscore', 'oae.core'], function($, _, oae) {
         // collabdocs have a special editor role
         if (contentProfile.resourceSubType === 'collabdoc') {
             widgetData.roles.editor = oae.api.i18n.translate('__MSG__CAN_EDIT__');
-            widgetData.defaultRole = 'editor';
+            widgetData.defaultRole = 'manager';
         }
         return widgetData;
     };
--- a/ui/js/content.js
+++ b/ui/js/content.js
@@ -294,19 +294,19 @@ require(['jquery', 'underscore', 'oae.core'], function($, _, oae) {
             'contextProfile': contentProfile,
             'messages': getManageAccessMessages(),
             'defaultRole': 'viewer',
-            'roles': {
-                'viewer': oae.api.i18n.translate('__MSG__CAN_VIEW__'),
-                'manager': oae.api.i18n.translate('__MSG__CAN_MANAGE__')
-            },
+            'roles': [
+                {'id': 'viewer', 'name': oae.api.i18n.translate('__MSG__CAN_VIEW__')},
+                {'id': 'manager', 'name': oae.api.i18n.translate('__MSG__CAN_MANAGE__')}
+            ],
             'api': {
                 'getMembersURL': '/api/content/'+ contentProfile.id + '/members',
                 'setMembers': oae.api.content.updateMembers,
                 'setVisibility': oae.api.content.updateContent
             }
         };
-        // collabdocs have a special editor role
+        // Collabdocs have a special editor role
         if (contentProfile.resourceSubType === 'collabdoc') {
-            widgetData.roles.editor = oae.api.i18n.translate('__MSG__CAN_EDIT__');
+            widgetData.roles.splice(1, 0, {'id': 'editor', 'name': oae.api.i18n.translate('__MSG__CAN_EDIT__')});
             widgetData.defaultRole = 'manager';
         }
         return widgetData;
--- a/ui/js/content.js
+++ b/ui/js/content.js
@@ -290,10 +290,10 @@ require(['jquery', 'underscore', 'oae.core'], function($, _, oae) {
      * @see manageaccess#initManageAccess
      */
     var getManageAccessData = function() {
-        return {
+        var widgetData = {
             'contextProfile': contentProfile,
             'messages': getManageAccessMessages(),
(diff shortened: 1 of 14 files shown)


stuartf:
I thought editors could only give others viewer rights, are they supposed to be able to add more editors as well?

nicolaasmatthijs:
@stuartf : Yes, viewers should be able to add both editors and viewers.

stuartf:
OK, but they can't remove viewers and editors correct?

nicolaasmatthijs:
@stuartf : They can only share the document (and decide whether those people should be editors or viewers in the meantime). They should not be able to update or remove roles on the document.

stuartf:
OK, but if someone is already a viewer and the editor shares it with that person and picks editor should that be an error, or should the role get updated? What about if the person is already an editor and another editor shares it to him and picks viewer is that an error or does he lose editor status?

nicolaasmatthijs:
What do we currently do when you share a file with a user that is already a manager? My inclination would be to just ignore the request if the role is less than that user's current role.

stuartf:
currently if you share and search for another manager and set it to `can view` they get downgraded to a viewer, but I think it makes sense for editors to not be able to take away access only grant it.

nicolaasmatthijs:
@stuartf : But that would only happen when the user sharing is a manager of the content, right? Or does it also do that when you're not the manager of the item?

stuartf:
Currently as a viewer of a collabdoc I don't see a share button.

nicolaasmatthijs:
There should be one in the clip, at least if the collabdoc is not private.

stuartf:
Ah, that was the problem, my collabdoc was private. Viewer sharing with manager results in the manager remaining a manager.

stuartf:
@nicolaasmatthijs I'm trying to remember what we decided to do here. IIRC it was to make editors able to share (add viewers) only, and set the default role for adding to a collabdoc back to manager. Is that right?

stuartf:
This is ready for review

ui/js/content.js
```diff
@@ -304,6 +304,12 @@ require(['jquery', 'underscore', 'oae.core'], function($, _, oae) {
                 'setVisibility': oae.api.content.updateContent
             }
         };
+        // collabdocs have a special editor role
```
nicolaasmatthijs: Should start with capital case

ui/js/content.js
```diff
@@ -304,6 +304,12 @@ require(['jquery', 'underscore', 'oae.core'], function($, _, oae) {
...
+        if (contentProfile.resourceSubType === 'collabdoc') {
+            widgetData.roles.editor = oae.api.i18n.translate('__MSG__CAN_EDIT__');
```
nicolaasmatthijs: The editor role should be displayed in between the viewer and manager role

ui/js/content.js
```diff
@@ -290,10 +290,10 @@ require(['jquery', 'underscore', 'oae.core'], function($, _, oae) {
      * @see manageaccess#initManageAccess
```
nicolaasmatthijs: When sharing a collaborative doc as an editor, there should be a dropdown in the "share" overlay (similar to manage access) where you can specify the role (can edit or can view)

nicolaasmatthijs:
Finished review. Back to @stuartf for follow-up.

nicolaasmatthijs:
@stuartf: Would you be able to look into the follow-up for this?

ui/js/content.js
stuartf: ECMAScript standard says an Object is an `unordered collection of properties`. Looking at how we use this `roles` needs to be an array of Objects.

ui/js/content.js
stuartf: I thought we decided to have editors only able to give `viewer` roles to others and leave `manager` as the default role.

nicolaasmatthijs: No, we decided that editors can add both other editors and viewers, so they are able to invite other people to work on the document. `manager` should indeed be the default for groups and users added during creation.

stuartf:
ready for more review

nicolaasmatthijs:
@stuartf : There's still some outstanding review here around sharing a document as an editor

nicolaasmatthijs:
@stuartf: How close to completion is the outstanding follow-up?

ui/js/content.js
```diff
@@ -290,20 +290,26 @@ require(['jquery', 'underscore', 'oae.core'], function($, _, oae) {
...
+        var widgetData = {
             'contextProfile': contentProfile,
             'messages': getManageAccessMessages(),
-            'defaultRole': contentProfile.resourceSubType === 'collabdoc' ? 'manager' : 'viewer',
-            'roles': {
-                'viewer': oae.api.i18n.translate('__MSG__CAN_VIEW__'),
-                'manager': oae.api.i18n.translate('__MSG__CAN_MANAGE__')
-            },
+            'defaultRole': 'viewer',
+            'roles': [
```
nicolaasmatthijs: This should have been done for groups, etc. as well. I'll see if I can push a fix in time for the bash

nicolaasmatthijs:
Followed up and merged

PR #3936, opened 2015-01-29
(A11y severity 3) Close notifications dropdown on loss of focus

sathomas:
When a keyboard user navigates out of the dropdown box, it remains open blocking the content below it from view. This can be difficult for sighted keyboard users as they navigate through elements hidden by the box. Ensure that the dropdown closes once a keyboard user navigates out of it.


sathomas:
to @nicolaasmatthijs for review. this fix is tricky enough to warrant extra care for the review

PR #3975, opened 2015-02-19
Allow tiles to show multiple lines of the item name

nicolaasmatthijs:
Improved item name readability: allow tiles to show multiple lines of the item name

### DIFF
--- a/node_modules/oae-core/lhnavigation/js/lhnavigation.js
+++ b/node_modules/oae-core/lhnavigation/js/lhnavigation.js
@@ -59,6 +59,9 @@ define(['jquery', 'oae.core', 'jquery.history'], function($, oae) {
             var $cachedPage = $('div[data-page="' + selectedPage.id + '"]', $pageContainer);
             if ($cachedPage.length > 0) {
                 $cachedPage.show();
+
+                // Re-apply multi-line threedotting to the tile titles
+                $(".oae-tile h3").trigger("update");
             } else {
                 $pageContainer.append(oae.api.util.template().render($('#lhnavigation-page-template'), {
                     'selectedPage': selectedPage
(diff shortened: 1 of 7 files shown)


nicolaasmatthijs:
Assigning to @sathomas . @sathomas : Would you have sometime to look into this?

sathomas:
I think Monday might be good, though I guess that depends on how Clay's presentation goes. (Well, I should have time before and during the presentation, at any rate.)

node_modules/oae-core/lhnavigation/js/lhnavigation.js
```diff
@@ -59,6 +59,9 @@ define(['jquery', 'oae.core', 'jquery.history'], function($, oae) {
             var $cachedPage = $('div[data-page="' + selectedPage.id + '"]', $pageContainer);
             if ($cachedPage.length > 0) {
                 $cachedPage.show();
+
+                // Re-apply multi-line threedotting to the tile titles
+                $(".oae-tile h3").trigger("update");
```
sathomas: Should those be single quotation marks instead of double quotation marks?

shared/oae/js/jquery-plugins/jquery.infinitescroll.js
```diff
@@ -263,6 +263,11 @@ define(['jquery', 'underscore', 'oae.api.util', 'oae.api.i18n', 'oae.api.l10n'],
 
                 // Apply timeago to the `oae-timeago` elements in the output container
                 oaeL10n.timeAgo($listContainer);
+
+                // Apply multi-line threedotting to the tile titles
+                $(".oae-tile h3").dotdotdot({
+                    watch: "window"
```
sathomas: same question on both of these line?

sathomas:
Looks good to be, but I left a couple of questions for @nicolaasmatthijs re: OAE style practice. (Otherwise, it's good to merge.)

sathomas:
As an FYI, in webkit it is possible to do multi-line ellipses with CSS only using the `-webkit-line-clamp` property. Doesn't work in Firefox or IE, of course. (Sure would be handy if it did.)

nicolaasmatthijs:
Followed up on review. Back to @sathomas

PR #3980, opened 2015-03-03
Weird blue border around folder selection

nicolaasmatthijs:
I have 1 folder shared with me.

I went into a collaborative document and selected to "Add to Folder". When I select the folder to which I want to add it, it gets a weird asymmetric blue border. Notice that I have switched focus to the "My Library" tab, which indicates that the blue border is *not* the browser's focus border.

screen shot 2015-02-26 at 11 27 50 am

Chrome 40.0.2214.111 Mac OSX

### DIFF
--- a/shared/oae/css/oae.components.css
+++ b/shared/oae/css/oae.components.css
@@ -50,10 +50,16 @@ a.oae-plain-link:hover {
  * Style that can be used to fake focus on an element. This can for example
  * be used when a hidden element needs to provide visual feedback on a different element.
  */
+
 .oae-focus {
     outline: thin dotted #333;
-    outline: 5px auto -webkit-focus-ring-color;
     outline-offset: -2px;
+}
+
+a.oae-focus,
+.btn.oae-focus,
+input.oae-focus {
+    outline: 5px auto -webkit-focus-ring-color;
 }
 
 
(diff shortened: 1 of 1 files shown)


nicolaasmatthijs:
Attached a fix. Assigning to @mrvisser for review

nicolaasmatthijs:
Note that the blue item is supposed to indicate the selected item, so it is expected to remain blue when you tab away from it. However, the asymmetry definitely was a visual issue.

shared/oae/css/oae.components.css
```diff
@@ -50,12 +50,18 @@ a.oae-plain-link:hover {
...
  */
+
 .oae-focus {
     outline: thin dotted #333;
-    outline: 5px auto -webkit-focus-ring-color;
     outline-offset: -2px;
 }
 
+a.oae-focus,
+.btn.oae-focus,
```
mrvisser: Do we already have a convention to ignore the `.` when doing alphabetical ordering? It looks funny to me to split the tagName selectors with a className selector. I think I'd prefer:

```css
a.oae-focus,
input.oae-focus,
.btn.oae-focus {
   ...
}
```

mrvisser:
Very minor feedback request, back to @nicolaasmatthijs

nicolaasmatthijs:
Followed up on review. Back to @mrvisser for merging.

PR #3984, opened 2015-03-11
Wait for body to be visible before applying $.dotdotdot

mrvisser:
There was a race condition where the $.dotdotdot plugin gets invoked while the body is not made visible yet. This change re-uses another area of `search.js` that seems to be listening for the same quirk.

Since `$('body').show()` seems to be in nature asynchronous, so it would be good to take some kind of event-driven approach to better trigger when the widgets have loaded.

### DIFF
--- a/shared/oae/api/oae.api.js
+++ b/shared/oae/api/oae.api.js
@@ -129,7 +129,7 @@ define(['oae.api.admin', 'oae.api.authentication', 'oae.api.config', 'oae.api.co
                                     oae.api.widget.loadWidgets(null, null, null, function() {
                                         // We can show the body as internationalization and
                                         // initial widget loading have finished
-                                        $('body').show();
+                                        $('body').css('visibility', 'visible');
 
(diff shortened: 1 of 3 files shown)


mrvisser:
Assigning to @simong for review

nicolaasmatthijs:
Rather than waiting for the body to become visible, I wonder if we would be better off changing the default visibility of the body from `display: none` to `visibility: hidden` and change `oae.api.js` to do `$('body').css('visibility', 'visible');` instead of `$('body').show()`. I *think* that should allow all calculations to just work whilst maintaining a smooth page load.

nicolaasmatthijs:
Left an idea for an alternative approach. Back to @mrvisser

mrvisser:
Good call, didn't realize using `visibility` resulted in that stuff being calculated but `display` didn't. I've made the change and it seems to work.

Back to @nicolaasmatthijs

nicolaasmatthijs:
Excellent! Merged

Issue #3985, opened 2015-03-12
Follow up on #3983


nicolaasmatthijs:
The following regressions were found when testing the fixes for https://github.com/oaeproject/3akai-ux/pull/3983/files:

- On Chrome, the jeditable field jumps down by 1 pixel when switching to edit mode
- The edit version of the jeditable field doesn't stretch as far as the view mode

![screen shot 2015-03-11 at 22 02 01](https://cloud.githubusercontent.com/assets/109850/6612674/f46c5580-c83d-11e4-80d8-85cd46d61ce1.png)

nicolaasmatthijs:
Assigning to @mrvisser

simong:
FWIW, This happens in production as well and isn't necessarily a regression from #3983

mrvisser:
Yea I found it happened on production as well which is why I didn't wrestle with it more to fix that one pixel on Chrome. I can keep looking at it but it's not going to be a case of just mingling a couple margins, I'll have to do some experimentation and hopefully something works.

nicolaasmatthijs:
Yeah, both of these happen on production as well. @mrvisser : Do you know how difficult it would be to only fix `The edit version of the jeditable field doesn't stretch as far as the view mode` ?

mrvisser:
@nicolaasmatthijs not at the moment, I'll look into it

mrvisser:
Note that this will not be possible with IE9, but we can make the `h3` for the item name 100% on other browser without causing wrapping, then give the `input` for the editable field a width of `100%`. But we should only let the `h3` be full width when it is editable, or else the fly-overs look funny.

mrvisser:
Closing in favour of #3988

PR #3988, opened 2015-03-13
(Issue #3985) Fix uneven width on jeditable list items

mrvisser:
I've fixed the jeditable width issue for all non-ie9 browsers. For IE9 the fluid width of the text field was always an issue, which is why we wound up with the fixed widths in the first place. This change makes it so that all modern browsers can have fluid width text fields that are the full width of the list item.

### DIFF
--- a/node_modules/oae-core/createlink/css/createlink.css
+++ b/node_modules/oae-core/createlink/css/createlink.css
@@ -29,11 +29,15 @@
     margin: 12px 0 0;
 }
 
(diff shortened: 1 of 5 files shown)


mrvisser:
Assigning to @nicolaasmatthijs for review

sallakarppinen:
Closing this in favour of a larger clean-up to remove IE9 specific CSS in issue #4118 .

PR #3997, opened 2015-03-19
(PrincipalsDelete) Add delete button for groups

mrvisser:
Adds a delete button to leverage the back-end group delete work made available in https://github.com/oaeproject/Hilary/pull/1149


shared/oae/api/oae.api.group.js
```diff
@@ -280,4 +280,32 @@ define(['exports', 'jquery', 'underscore', 'oae.api.util'], function(exports, $,
             }
         });
     };
+
+    /**
+     * Delete a group as the currently authenticated user.
```
nicolaasmatthijs: We can just say `Delete a group` here. The join and leave function just have the `as the currently authenticated user` bit as it makes more sense there

nicolaasmatthijs:
Left one small bit of feedback. Ready to merge after that. Back to @mrvisser

nicolaasmatthijs:
We usually have a button in the library as well to allow deleting multiple items at once. However, I'm thinking that we might be better off not offering that for groups.

mrvisser:
Feedback addressed, back to @nicolaasmatthijs . I suppose this should go `On Hold` until the back-end side of this is ready for merge as well.

nicolaasmatthijs:
Ready for merging. Putting on hold until oaeproject/Hilary#1149 is ready to merge

mrvisser:
I totally forgot about the library. But glad we don't need it :)

PR #3998, opened 2015-03-19
Multi-line threedotting issues in modals

nicolaasmatthijs:
screen shot 2015-03-19 at 10 11 24


nicolaasmatthijs:
Attached a fix. Over to @mrvisser for review

mrvisser:
@nicolaasmatthijs this should happen for `foldershared.js` as well I think

nicolaasmatthijs:
That escalated quickly. Back to @mrvisser

node_modules/oae-core/sharedwith/tests/sharedwith.js
```diff
@@ -0,0 +1,139 @@
+/*!
+ * Copyright 2014 Apereo Foundation (AF) Licensed under the
```
mrvisser: `2015` :)

nicolaasmatthijs:
Followed up on comment. Back to @mrvisser

node_modules/oae-core/sharedwith/js/sharedwith.js
```diff
@@ -57,24 +58,27 @@ define(['jquery', 'oae.core'], function($, oae) {
         };
 
         /**
-         * Initialize the discussionshared modal dialog
+         * Initialize the sharedwith modal dialog
          */
-        var setUpDiscussionShared = function() {
-            $(document).on('click', '.oae-trigger-discussionshared', function(ev, data) {
+        var setUpsharedwith = function() {
```
mrvisser: `setUpSharedWith`, I'll fix

node_modules/oae-core/sharedwith/js/sharedwith.js
```diff
@@ -26,21 +26,22 @@ define(['jquery', 'oae.core'], function($, oae) {
         /**
          * Initialize a new infinite scroll container that fetches the members
          *
-         * @param  {Discussion}    discussionProfile    Discussion for which the members should be rendered
+         * @param  {Content|Discussion|Folder}    contextProfile    Entity for which the members should be rendered
```
mrvisser: We don't really have a reliable process to come back and update this for each resource type that gets added in the future. We should probably just specify `Object` and indicate it accepts any resource model with a `resourceType`, `id` and a members list

mrvisser: I'll fix this too

PR #3999, opened 2015-03-27
Group profiles

mrvisser:


mrvisser:
I still need to do a self-review and cross-browser testing, especially IE9/10/11

node_modules/oae-core/members/js/members.js
```diff
@@ -46,7 +46,7 @@ define(['jquery', 'oae.core', 'jquery.history'], function($, oae) {
 
             var url = '/api/group/' + widgetData.context.id + '/members';
             if (query) {
-                url = '/api/search/members/'  + widgetData.context.id;
+                url = '/api/search/group-members-library/'  + widgetData.context.id;
```
mrvisser: Needs to change to `members-library`

node_modules/oae-core/activity/js/activity.js
```diff
@@ -42,7 +42,8 @@ define(['jquery', 'underscore', 'oae.core', 'activityadapter'], function($, _, o
                     'additionalClasses': 'col-xs-12 col-lg-6'
                 },
                 'displayOptionsSmall': {
-                    'additionalClasses': 'col-xs-6 col-sm-4 col-md-3 col-lg-2'
+                    'additionalClasses': (widgetData.context.resourceType === 'group') ?
+                        'col-xs-4 col-sm-3 col-md-4 col-lg-3' : 'col-xs-6 col-sm-4 col-md-3 col-lg-2'
```
mrvisser: Ideally we have a better strategy for this. I'm not sure if I should just allow these col classes to come in as the activity widget settings, or if we should have named orientations. I'm leaning towards widget settings, in which case we probably want one for the `displayOptionsSmall` and the `displayOptionsLarge`

node_modules/oae-core/members/js/members.js
sathomas: Not sure of the context, but are we changing the API endpoint from `/api/search/members/g:xx:xxx` to `/api/search/members-library/g:xx:xxx` or are we just adding a new endpoint?

If the former, that's a problem. The STEM incubator site relies on the existing API endpoint. If that endpoint goes away, then the site breaks.

If this is just a copy-and-paste issue and the existing API endpoint is sticking around, then never mind.

mrvisser: "members" search is going away in place of "members-library", which is effectively the same, but more permissive than what the old members search was. How difficult would it be for us to coordinate this change? If there absolutely needs to be a transition period, I suppose we could do that.

khomotso: As a general rule, does the project have a policy on backwards compatibility of API changes, or breaking API changes more generally? We're getting into a situation where our production deployment depends on the stability of the APIs as much as it does the stability of the rest of the platform.

mrvisser: From a product development standpoint we have a general convention that breaking changes require at least a minor version bump. So those with their own OAE deployments should be wary of API changes both internal and external and test accordingly prior to releasing a minor version upgrade.

From a hosting provider stand-point, we do not have a policy on releasing non-backwards-compatible API changes to production. It seems we should at least get acknowledgement from GeorgiaTech before releasing such a change, which is simple enough. I'm not aware of any other production API consumers at this point, but we can probably get an idea of external usage from our web analytics, actually.

@nicolaasmatthijs do you have any idea off hand how much usage comes from what looks like direct API access, if any?

khomotso: Special consideration for Georgia Tech would of course be appreciated, but
the project is probably better served by a standard deprecation process
which anyone could follow, to which Georgia Tech could be asked to refer to
along with everyone else. For the moment, Georgia Tech is wearing the hats
of two distinct stakeholders, but in a few months those hats will be on
separate heads again.

I haven't yet looked at other API policies for comparable projects as a
point of comparison, but will do so.

nicolaasmatthijs:
Found the following items in a functional review:

- The column layout on the activity page can be switched to 9-3 instead of 8-4 on larger screens
- The color of the pencil icon should be lighter as per the designs
- `Add an introduction to this group ` can be reworded as `Add a description to this group`
- `Add a description to this group` should be italic as per the designs
- When clicking the pencil icon or `Add a description to this group`, the description field should be focussed in the edit group modal
- I get a 400 error when trying to search the members library:

```
{"code":400,"msg":"Search \"group-members-library\" is not a valid search type."}
```

- The visibility icon in the clip seems messed up:

screen shot 2015-03-31 at 23 28 07

- The icons in the upload widget seem to be messed up:

screen shot 2015-03-31 at 23 28 43

- The activity feed has been coded in such a way that it will only show the first line of the items involved in an activity. The remaining items are hidden behind a `Show all` link. Whilst the implementation here ensure that the size of the tiles is adjusted correctly based on how much space the activity widget has available, it causes more than 1 line of results to be displayed:

screen shot 2015-03-31 at 23 31 05

- The spacing below and above the member thumbnails isn't quite right. The spacing above the thumbnails should be increased by 5-10px, whilst the spacing below the thumbnails should be decreased by 5-10px.

nicolaasmatthijs:
Back to @mrvisser for follow-up

mrvisser:
@nicolaasmatthijs just FYI, the review comments that are bugs you found I already fixed from my testing in https://github.com/mrvisser/3akai-ux/commit/0192ed0d860d9739b5eb6251c3407288ee327739 . Though the `members-library` one I don't know how it slipped through :/ There are other more subtle style things that are relevant in here though that I'll fix today.

mrvisser:
> Though the members-library one I don't know how it slipped through :/

Hah, I dropped myself that review comment above and it manifested into a large discussion about backward compatibility :)

mrvisser:
> The activity feed has been coded in such a way that it will only show the first line of the items involved in an activity. The remaining items are hidden behind a Show all link. Whilst the implementation here ensure that the size of the tiles is adjusted correctly based on how much space the activity widget has available, it causes more than 1 line of results to be displayed:

I can't seem to reproduce this on the latest version, can you try and update/clear cache etc... and see if you still see this?

mrvisser:
Followed up and re-implemented thumbnail "Show all", back to @nicolaasmatthijs

node_modules/oae-core/about/about.html
```diff
@@ -0,0 +1,48 @@
+<!-- CSS -->
```
nicolaasmatthijs: The `about` widget should be renamed to `aboutgroup`, given that it will only work for groups as is

node_modules/oae-core/about/about.html
```diff
@@ -0,0 +1,48 @@
...
+<link rel="stylesheet" type="text/css" href="css/about.css" />
+
+<div class="about-widget">
+    <div id="about-header"><!-- --></div>
```
nicolaasmatthijs: We usually suffix elements that will be filled with template content with `-container`

node_modules/oae-core/about/about.html
```diff
@@ -0,0 +1,48 @@
...
+    <div id="about-members"><!-- --></div>
+</div>
+
+<div id="about-header-template"><!--
+    ${header('__MSG__ABOUT__', action)}
+--></div>
+
+<div id="about-description-template"><!--
+    {if contextProfile.description}
+        <div class="about-description-content">
```
nicolaasmatthijs: In which case this can just become `aboutgroup-description`

node_modules/oae-core/about/about.html
```diff
@@ -0,0 +1,48 @@
...
+            ${oae.api.util.security().encodeForHTMLWithLinks(contextProfile.description).replace(/\n/g, '<br/>')}
```
nicolaasmatthijs: I've recently discovered that you can use `white-space: pre-wrap;` or `white-space: pre-line;` to ensure that line breaks are preserved, removing the need to do the replace. I'd be in favor of using that here and whilst we're at it also introduce it in the few other places where we're doing this

http://stackoverflow.com/questions/19684708/preserve-line-breaks-in-angularjs

node_modules/oae-core/about/about.html
```diff
@@ -0,0 +1,48 @@
...
+        </div>
+    {else}
+        <div class="about-description-empty">
+            {if contextProfile.isManager}
+                <a class="group-trigger-editgroup" data-editgroup-focus="#editgroup-description" href="#" title="__MSG__ADD_AN_INTRODUCTION_TO_THIS_GROUP__">
```
nicolaasmatthijs: Given that this doesn't link to a new page, this should be a button styled as a link

node_modules/oae-core/about/about.html
```diff
@@ -0,0 +1,48 @@
...
+                    __MSG__ADD_AN_INTRODUCTION_TO_THIS_GROUP__
```
nicolaasmatthijs: This should be `ADD_A_DESCRIPTION_TO_THIS_GROUP`. The title attribute doesn't need to be set as it doesn't provide any extra information on top of the element's text

node_modules/oae-core/about/about.html
```diff
@@ -0,0 +1,48 @@
...
+                </a>
+            {else}
+                __MSG__NO_INTRODUCTION_PROVIDED_YET__
```
nicolaasmatthijs: `NO_DESCRIPTION_PROVIDED_YET`

node_modules/oae-core/about/about.html
```diff
@@ -0,0 +1,48 @@
...
+            {/if}
+        </div>
+    {/if}
+--></div>
+
+<div id="about-members-template"><!--
+    <div class="about-members">
+        {for thumbnail in thumbnails}
```
nicolaasmatthijs: I'd prefer to do `{for member in members}` here

node_modules/oae-core/about/about.html
```diff
@@ -0,0 +1,48 @@
...
+            ${renderThumbnail(thumbnail.entity, thumbnail.displayOptions)}
+        {/for}
+    </div>
+    <div class="about-show-all">
+        {if thumbnails.length > 0}
```
nicolaasmatthijs: Should this just be wrapped around the entire template?

node_modules/oae-core/about/about.html
```diff
@@ -0,0 +1,48 @@
...
+            <a href="${contextProfile.profilePath}/members" title="__MSG__SHOW_ALL_MEMBERS__">
```
nicolaasmatthijs: No need to `title` attribute here

node_modules/oae-core/about/bundles/default.properties
```diff
@@ -0,0 +1,3 @@
+ADD_AN_INTRODUCTION_TO_THIS_GROUP=Add a description to this group
```
nicolaasmatthijs: Name of key should be update to have `DESCRIPTION` instead of `INTRODUCTION`

node_modules/oae-core/about/bundles/default.properties
```diff
@@ -0,0 +1,3 @@
...
+NO_INTRODUCTION_PROVIDED_YET=No introduction provided yet
```
nicolaasmatthijs: Should be `DESCRIPTION` / `description`

node_modules/oae-core/about/css/about.css
```diff
@@ -0,0 +1,62 @@
+
```
nicolaasmatthijs: Needs license header

node_modules/oae-core/about/css/about.css
```diff
@@ -0,0 +1,62 @@
...
+#about-description .about-description-empty,
+#about-description .about-description-empty a {
+    color: #888;
```
nicolaasmatthijs: I'd prefer not to hardcode this color here, but it's probably not worth adding a skin value for it, so no action needed

node_modules/oae-core/about/css/about.css
```diff
@@ -0,0 +1,62 @@
+
+#about-description .about-description-empty,
```
nicolaasmatthijs: All rules should be prefixed with `.about-widget` (or `.aboutgroup-widget` once renamed)

node_modules/oae-core/about/css/about.css
```diff
@@ -0,0 +1,62 @@
...
+#about-description .about-description-empty a {
+    font-style: italic;
+    text-decoration: underline;
+}
+
+#about-members {
+    margin-top: 35px;
+}
+
+#about-members > div.about-show-all {
```
nicolaasmatthijs: No need for `div` part here

node_modules/oae-core/about/css/about.css
```diff
@@ -0,0 +1,62 @@
...
+    margin-top: 15px;
+}
+
+#about-members .about-members {
+    display: -webkit-box;
```
nicolaasmatthijs: I'm pretty sure we don't need any of these `display` and `flex-wrap` properties. Something like a `display: inline-block` should be sufficient

node_modules/oae-core/about/css/about.css
```diff
@@ -0,0 +1,62 @@
...
+    -webkit-flex-wrap: wrap;
+        -ms-flex-wrap: wrap;
+            flex-wrap: wrap;
+}
+
+#about-members .about-members div.oae-thumbnail {
+    border: none;
+    height: 60px;
+    width: 60px;
+    -webkit-box-flex: none;
```
nicolaasmatthijs: Again, I don't think we need to set any of these.

node_modules/oae-core/about/css/about.css
```diff
@@ -0,0 +1,62 @@
...
+/* Clear the border radius on base oae-thumbnail style */
+#about-members .about-members div.oae-thumbnail,
+#about-members .about-members div.oae-thumbnail a,
+#about-members .about-members div.oae-thumbnail div {
+    -webkit-border-radius: 0;
+       -moz-border-radius: 0;
+            border-radius: 0;
+}
+
+/* IE9 adds 5px of space after the images for no reason. Give the individual thumbnails a negative
```
nicolaasmatthijs: I'm pretty sure this behavior will go away once we remove all of the `box-flex` and other `flex` things in favor of something like `inline-block`

node_modules/oae-core/about/css/about.css
```diff
@@ -0,0 +1,62 @@
...
+        -webkit-flex: none;
+            -ms-flex: none;
+                flex: none;
+}
+
+/* Clear the border radius on base oae-thumbnail style */
+#about-members .about-members div.oae-thumbnail,
+#about-members .about-members div.oae-thumbnail a,
+#about-members .about-members div.oae-thumbnail div {
+    -webkit-border-radius: 0;
```
nicolaasmatthijs: According to http://caniuse.com/#feat=border-radius, it should be safe to now just use `border-radius`. If we have any other place where we're still setting all of the unnecessary prefixed versions, we can remove those as well.

node_modules/oae-core/about/js/about.js
```diff
@@ -0,0 +1,94 @@
...
+        /**
+         * Render the header of the about widget
+         */
+        var setUpAboutHeader = function() {
+            var action = null;
+
+            if (contextProfile.isManager) {
+                action = {
+                    'icon': 'fa-pencil',
+                    'trigger': 'group-trigger-editgroup',
```
nicolaasmatthijs: I think this widget was originally coded to work with all entities, but these things bind it to groups only

node_modules/oae-core/about/js/about.js
```diff
@@ -0,0 +1,94 @@
...
+            }
+
+            oae.api.util.template().render($('#about-header-template', $rootel), {'action': action}, $('#about-header', $rootel));
+        };
+
+        /**
+         * Render the profile description of the about widget
+         */
+        var setUpAboutDescription = function() {
+            oae.api.util.template().render($('#about-description-template', $rootel), {'contextProfile': contextProfile}, $('#about-description', $rootel));
```
nicolaasmatthijs: You could just pass in `contextProfile` and do `if (description) {` in the template. Whichever you find more readable

node_modules/oae-core/about/js/about.js
```diff
@@ -0,0 +1,94 @@
...
+                    .value()
+            };
+
+            oae.api.util.template().render($('#about-members-template', $rootel), data, $('#about-members', $rootel));
+        };
+
+        /**
+         * Re-render the about description content when the group profile updates
+         */
+        $(document).on('oae.about.profile-update', function(ev, updatedContextProfile) {
```
nicolaasmatthijs: It would be good to not make this an event specific to this widget that then needs to be added in other widgets. A more general event like `oae.context.update` (cfr `'oae.context.get'`) would be more appropriate

node_modules/oae-core/activity/css/activity.css
```diff
@@ -75,78 +85,21 @@
...
     margin-bottom: -15px;
     margin-top: 20px;
 }
 
-.activity-widget .activity-preview-container ul.oae-list[data-items="1"] + .activity-show-all-toggle-container,
-.activity-widget .activity-preview-container ul.oae-list[data-items="2"] + .activity-show-all-toggle-container {
+.activity-widget .activity-preview-container .activity-show-all-toggle-container,
+.activity-widget .activity-preview-container.activity-show-all .activity-show-all-toggle-container,
+.activity-widget .activity-preview-container.activity-show-all.activity-overflow .activity-show-all-toggle-container {
     display: none;
```
nicolaasmatthijs: Can we not just use `hide` here instead of defining CSS rules?

nicolaasmatthijs: Even if we can't, the class structure is a bit hard to decipher.

node_modules/oae-core/activity/js/activity.js
```diff
@@ -123,7 +124,37 @@ define(['jquery', 'underscore', 'oae.core', 'activityadapter'], function($, _, o
                 'limit': 10
             }, '#activity-items-template', {
                 'postProcessor': processActivities,
-                'emptyListProcessor': handleEmptyResultList
+                'emptyListProcessor': handleEmptyResultList,
+                'postRenderer': function(data, $lis) {
```
nicolaasmatthijs: Should rename `$lis` to something more meaningful

node_modules/oae-core/activity/js/activity.js
```diff
@@ -123,7 +124,37 @@ define(['jquery', 'underscore', 'oae.core', 'activityadapter'], function($, _, o
...
+                    $lis.find('ul.oae-list').each(function(i, ul) {
+                        var $ul = $(ul);
```
nicolaasmatthijs: `$ul = $(this);`. No need for `i` and `ul`

node_modules/oae-core/editgroup/js/editgroup.js
```diff
@@ -124,8 +134,8 @@ define(['jquery', 'oae.core'], function($, oae) {
             });
 
             $('#editgroup-modal', $rootel).on('shown.bs.modal', function () {
-                // Set focus to the group name field
-                $('#editgroup-name', $rootel).focus();
+                // Set focus to the specified field, defaulting to the name field
```
nicolaasmatthijs: `group name field`

node_modules/oae-core/editgroup/js/editgroup.js
```diff
@@ -101,6 +110,7 @@ define(['jquery', 'oae.core'], function($, oae) {
          */
         var setUpEditGroupModal = function() {
             $(document).on('click', '.group-trigger-editgroup', function() {
+                focusSelector = $(this).attr('data-editgroup-focus');
```
nicolaasmatthijs: You could just do `|| #editgroup-name` here

shared/oae/css/oae.components.css
```diff
@@ -1009,7 +1014,6 @@ div.oae-thumbnail i.fa {
 .oae-tile,
 .oae-list.oae-list-grid > li.oae-list-actions {
     position: relative;
-    overflow: hidden;
```
nicolaasmatthijs: Do you know what this was here for?

shared/oae/js/jquery-plugins/jquery.infinitescroll.js
```diff
@@ -36,7 +36,7 @@ define(['jquery', 'underscore', 'oae.api.util', 'oae.api.i18n', 'oae.api.l10n'],
...
      * @param  {Function}                          [options.emptyListProcessor]    Function that will be executed when the rendered list doesn't have any elements.
      * @param  {Function}                          [options.postProcessor]         Function used to transform the search results before rendering the template. This function will be called with a data parameter containing the retrieved data and should return the processed data
-     * @param  {Function}                          [options.postRenderer]          Function executed after the rendered HTML has been appended to the rendered list. The full retrieved server response will be passed into this function.
+     * @param  {Function}                          [options.postRenderer]          Function executed after the rendered HTML has been appended to the rendered list. The full retrieved server response will be passed into this function as the first parameter as well as jQuery object containing the template that was added to the document.
```
nicolaasmatthijs: `containing the template that was added to the document` should be renamed to something like `the rendered template`

shared/oae/macros/header.html
```diff
@@ -0,0 +1,19 @@
...
+{macro header(title, action)}
+    <div class="oae-header">
+        <div class="row">
+            <div class="oae-header-title col-xs-12 col-sm-10">
+                <h2>${title}</h2>
+            </div>
+            {if action}
+                <div class="oae-header-action hidden-xs col-sm-2">
+                    <button class="btn btn-link ${action.trigger}" ${%_.map(action.data, function(value, key) { return 'data-' + key + '="' + value + '"'; }).join(' ')%}>
+                        <i class="fa ${action.icon}"></i>
```
nicolaasmatthijs: These actions need descriptive text for accessibility (both a `title` and a `<span class="sr-only"></span>` within the `i`)

shared/oae/macros/list.html
```diff
@@ -200,13 +200,14 @@
...
+ * @param  {String}                                         [displayOptions.customImage]        URL to the custom entity profile picture that should be used. If not provided, this will fall back to the thumbnailUrl
+ * @param  {String}                                         [displayOptions.flyover]            The content to put in a fly-over when the mouse hovers over the thumbnail image
+ * @param  {String}                                         [displayOptions.linkTarget]         The target of the thumbnail link. One of `_blank`, `_self`, `_parent` or `_top`. By default, the link will open in the same window
+ * @param  {String}                                         [displayOptions.size]               If specified, this content will suffix a class `oae-thumbnail-<size>`. Core CSS has support for "medium" and "large", and by default the `oae-thumbnail` is styled with what we would consider "small". "large" is recommend for thumbnails larger than 100px, while "medium" is recommended for larger than 50px
```
nicolaasmatthijs: You're using `small` above, so it should at least be specified as an expected option

nicolaasmatthijs:

node_modules/oae-core/about/about.html
mrvisser: There's already an `about<resourceType>` widget for every resource type. There is nothing that strictly makes this for groups, except for some message labels that are in there for simplicity while group is the only consumer of the widget.

node_modules/oae-core/about/js/about.js
mrvisser: The intention going in was that it would be rather generic, but there turned to be more one-offs than I thought that made it not worth the trouble upfront given we don't have an immediate need for another resource type. I also can't call it `aboutgroup` because there already is one.

node_modules/oae-core/about/js/about.js
mrvisser: Isn't that what I'm doing here?

nicolaasmatthijs: No, you're passing in an object with `contextProfile` as the only key and the `contextProfile` object as the value of that key, which means that you have to do `contextProfile.description` in the template. You could just pass in `contextProfile` as the data object, so you can just do `if (description) {` in the template.

```
oae.api.util.template().render($('#about-description-template', $rootel), contextProfile, $('#about-description', $rootel));
```

node_modules/oae-core/activity/css/activity.css
mrvisser: Being able to handle these states with 3 CSS rules is actually far more simple than all the conditional checks that would have to go into the JS to do the same thing on resize and on click using `hide()`

node_modules/oae-core/about/js/about.js
nicolaasmatthijs: Should we just call it something like `groupprofile` ?

node_modules/oae-core/editgroup/js/editgroup.js
mrvisser: I'm accounting for the potential that the editgroup modal can be triggered by some means other than a `.group-trigger-editgroup` class. Is that worthwhile? If not, then I think I can follow through with this feedback and cleanup the `focusSelector = null;` handler above as well.

shared/oae/css/oae.components.css
mrvisser: I don't know why it was there, but I'm seeing borders being chopped off in the activity tile depending on the width of the screen because of this. I think they are one-off pixel issues in at least chrome due to %-based widths.

I can't think of any situation where we actually want to chop something out of the UI with `overflow: hidden`, can you? If we are, we *probably* want to do something else, so I was hoping it would come up in testing, but it hasn't so far.

mrvisser:
Back to @nicolaasmatthijs

node_modules/oae-core/editgroup/js/editgroup.js
nicolaasmatthijs: Yeah, that makes sense. No need to change the current approach.

ui/js/group.js
```diff
@@ -160,29 +160,40 @@ require(['jquery', 'oae.core'], function($, oae) {
...
+
+        // Show the activity and about widgets on the main page
+        lhNavPages.push({
+            'id': 'activity',
+            'title': oae.api.i18n.translate('__MSG__RECENT_ACTIVITY__'),
+            'icon': 'fa-tachometer',
+            'closeNav': true,
+            'layout': [
+                {
+                    'width': 'col-md-9',
```
nicolaasmatthijs: @mrvisser : Would you be able to change this to `col-md-8 col-lg-9` for the recent activity widget and `col-md-4 col-lg-3` for the group profile widget? The current implementation is slightly too narrow at a narrow md width.

node_modules/oae-core/groupprofile/bundles/default.properties
```diff
@@ -0,0 +1,3 @@
+ADD_A_DESCRIPTION_TO_THIS_GROUP = Add a description to this group
+NO_DESCRIPTION_PROVIDED_YET = No description provided yet
```
nicolaasmatthijs: @SamPeck : When no description is available for a non-manager, do you want this message to be shown or should we just show the members?

screen shot 2015-04-07 at 21 55 32

nicolaasmatthijs:
Left one more comment and a question for @SamPeck. After that, this should be good to be merged.

node_modules/oae-core/groupprofile/bundles/default.properties
SamPeck: Just show the members
nonmanageraboutempty

mrvisser: Noted!

nicolaasmatthijs: You can keep the "About" header though

SamPeck: I would recommend not to keep that header, only because "About" doesn't make sense if only member thumbnails display.

nicolaasmatthijs: Really? It feels like the profile pics are floating too much without any context. `About` as the header doesn't seem that weird to me.

SamPeck: The profile pics have context in the link below them.

mrvisser: Should the profile pics move upward now that there is no header, or do we keep a blank strip of space where the header would have been? Keeping in mind the activity header to the left.

SamPeck: If we decide to remove the 'about' header then yes the profile pics should move up

mrvisser: screen shot 2015-04-14 at 1 12 27 pm

mrvisser:
This is ready for more review. I have changed it to always show `About` header for now.

ui/js/group.js
nicolaasmatthijs: This still needs to be addressed

ui/js/group.js
```diff
@@ -160,29 +160,40 @@ require(['jquery', 'oae.core'], function($, oae) {
...
-                                }
+
+        // Show the activity and about widgets on the main page
+        lhNavPages.push({
+            'id': 'activity',
+            'title': oae.api.i18n.translate('__MSG__RECENT_ACTIVITY__'),
+            'icon': 'fa-tachometer',
+            'closeNav': true,
+            'layout': [
+                {
```
nicolaasmatthijs: On `sm` and `xs` resolutions, the right hand bar should be hidden. It should not jump below the activity widget.

screen shot 2015-04-20 at 19 51 58

node_modules/oae-core/groupprofile/groupprofile.html
```diff
@@ -0,0 +1,42 @@
...
+    ${header('__MSG__ABOUT__', action)}
+--></div>
+
+<div id="groupprofile-description-template"><!--
+    {if contextProfile.description}
+        <div class="groupprofile-description">${oae.api.util.security().encodeForHTMLWithLinks(contextProfile.description)}</div>
+    {elseif contextProfile.isManager}
+        <div class="groupprofile-description-empty">
+            <button class="btn btn-link group-trigger-editgroup" data-editgroup-focus="#editgroup-description">
+                __MSG__ADD_A_DESCRIPTION_TO_THIS_GROUP__
```
nicolaasmatthijs: The text in this button should jump to multiple lines when necessary:

screen shot 2015-04-20 at 19 51 45

node_modules/oae-core/groupprofile/groupprofile.html
```diff
@@ -0,0 +1,42 @@
...
+            </button>
+        </div>
+    {/if}
+--></div>
+
+<div id="groupprofile-members-template"><!--
+    {if members.length > 0}
+        <div class="groupprofile-members">
```
nicolaasmatthijs: When members are added/removed, the list of members in the groupprofile widget should be updated as well

node_modules/oae-core/groupprofile/css/groupprofile.css
```diff
@@ -0,0 +1,51 @@
...
+    color: #888;
+    font-style: italic;
+    padding: 0 0 0 1px;
+    text-decoration: underline;
+}
+
+.groupprofile-widget .groupprofile-members {
+    margin-bottom: 15px;
+    margin-left: 4px;
+    width: 180px;
```
nicolaasmatthijs: You can remove this line and just make the profile pictures use the available width

nicolaasmatthijs:
Finished final round of review on this. Should be good to merge after follow-up

node_modules/oae-core/groupprofile/css/groupprofile.css
mrvisser: Sure, but just to be clear, if you have 9 members showing, on desktop resolution you have one little lingering one (they go 4 wide)

mrvisser: Would you want to decrease to 8? That way you never have 1 lingering. It's only ever full, or you have an incomplete row of 2 (the widgets make it switch between 3-wide and 4-wide)

mrvisser:
Updated all review feedback up to this point.

My latest commit also fixes some pre-existing bugs related to group profile updates when access has changed for the current user (e.g., clip items change appropriately).

node_modules/oae-core/groupprofile/css/groupprofile.css
nicolaasmatthijs: @mrvisser : Yes, switching to 8 sounds reasonable.

nicolaasmatthijs:
Followed up on stylistic feedback and merged.

PR #4005, opened 2015-04-06
(MySearch) New search page

mrvisser:
This is built on GroupProfiles, which should be merged first, though could be separated if needed.

Depends on https://github.com/oaeproject/Hilary/pull/1161 in the backend.


mrvisser:
Some notes about the "switch" component:

* Can take any number of elements, but probably only feasible to have 2 when in tablet or smaller
* There is a `min-width` to keep the elements the same size for small amount of text, but if one element gets more text they will be asymmetric, I feel like that's ok
* When down to tablet size and smaller, horizontal padding is reduced to ensure we can get a bit more text than the english defaults (~23 chars at `font-size: 15px`)

ui/search.html
```diff
@@ -23,79 +23,111 @@
...
+            <div class="oae-main-content">
+                <!-- EVERYWHERE / MY STUFF ONLY -->
+                <div id="search-scope" class="oae-switchtab-container">
+                    <div class="oae-switchtab-control">
+                        <button class="oae-switchtab-switch" data-switchtab-id="all" data-switchtab-path="/search/all">
+                            <span class="oae-switchtab-switch-icon fa fa-globe"></span>
+                            <span class="oae-switchtab-switch-label">__MSG__EVERYWHERE__</span>
+                        </button>
+                        <button class="oae-switchtab-switch" data-switchtab-id="my" data-switchtab-path="/search/my">
+                            <span class="oae-switchtab-switch-icon fa fa-lock"></span>
```
mrvisser: I'm deviating from the main icons in that:

* I use the public / private icon for everything/my searches
* Rather than use a check-icon for the selected element, I just add colour to the element icon

I don't mind changing these, but just thought I'd suggest it.

shared/oae/css/oae.components.css
```diff
@@ -198,6 +194,86 @@ i.fa.large:before {
...
+
+/* Highlight the active "switch" by giving it a background */
+.oae-switchtab-container .oae-switchtab-control .oae-switchtab-switch.oae-switchtab-active {
+    background-color: #EEE;
+    color: #333;
+}
+
+/* Highlight the active "switch"'s icon with a color, how nice! */
+.oae-switchtab-container .oae-switchtab-control .oae-switchtab-switch.oae-switchtab-active .oae-switchtab-switch-icon {
+    color: #3481BE;
```
mrvisser: I'm not exactly sure what skin variable this would be congruent with, but my assumption is that this colour should actually be brand-able.

shared/oae/css/oae.components.css
```diff
@@ -154,9 +152,7 @@ i.fa-oae-public {
 i.fa-oae-private:before,
 i.fa-oae-loggedin:before,
 i.fa-oae-public:before {
-    -webkit-border-radius: 500px;
-       -moz-border-radius: 500px;
-            border-radius: 500px;
+    border-radius: 500px;
```
mrvisser: ಠ_ಠ

mrvisser:
Assigning to @nicolaasmatthijs for review

mrvisser:
Self-reviewed, but the search page title needs testing and work I think.

mrvisser:
I've fixed the browser title. This should be good for review.

node_modules/oae-core/topnavigation/js/topnavigation.js
```diff
@@ -136,7 +136,7 @@ define(['jquery', 'oae.core', 'activityadapter'], function($, oae, ActivityAdapt
                 // full control of the URL hash and adding one  into the URL ourself would interfere with that
                 // @see https://github.com/oaeproject/3akai-ux/issues/3872
                 query = query.replace(/#/g, '');
-                window.location = '/search/' + oae.api.util.security().encodeForURL(query);
+                window.location = '/search/all?q=' + oae.api.util.security().encodeForURL(query);
```
nicolaasmatthijs: For logged in users, searching from the top navigation can point to the "My Stuff" page by default.

shared/oae/api/oae.api.util.js
```diff
@@ -175,13 +175,25 @@ define(['exports', 'require', 'jquery', 'underscore', 'oae.api.config', 'markdow
...
                 // Translate and cache the macros. We require the i18n API here to avoid creating
                 // a cyclic dependency
                 var i18nAPI = require('oae.api.i18n');
-                globalMacros.push(i18nAPI.translate(listMacro));
-                globalMacros.push(i18nAPI.translate(headerMacro));
+
+                // All loaded macros will come back as the arguments of the callback. Dynamically
+                // add them all to the global macros list
+                _.chain(arguments).toArray().each(function(macro) {
+                    globalMacros.push(macro);
```
nicolaasmatthijs: We've lost the translation of i18n keys in these templates. I think that's why the list views are now showing '__MSG__GROUP__' instead of "Group"

shared/oae/bundles/ui/default.properties
```diff
@@ -450,7 +452,7 @@ QUOTED_NOT_FOUND = Stephen Hawking
 QUOTE_ACCESS_DENIED = A man has no ears for that to which experience has given him no access.
 QUOTE_NOT_FOUND = Nothing cannot exist forever.
 RECENT_ACTIVITY = Recent activity
-REFINE_SEARCH_BY_TYPE_COLON = Refine your search by type:
+REFINE_YOUR_SEARCH_COLON = Refine your search:
```
nicolaasmatthijs: You'll need to remove/update the keys in the other i18n files as well

shared/oae/css/oae.base.css
```diff
@@ -596,6 +596,18 @@ ul.as-list li.as-result-item.active {
     background-size: cover;
 }
 
+/* MAIN ACTION DIV */
+
+.oae-main-action {
+    margin: 35px 0 0 0;
```
nicolaasmatthijs: This can be `15px` in all scenarios

shared/oae/css/oae.base.css
```diff
@@ -596,6 +596,18 @@ ul.as-list li.as-result-item.active {
     background-size: cover;
 }
 
+/* MAIN ACTION DIV */
```
nicolaasmatthijs: Maybe add some additional docs on what this is intended to be used for?

shared/oae/css/oae.components.css
```diff
@@ -194,6 +194,85 @@ i.fa.large:before {
 }
 
 
+/****************
```
nicolaasmatthijs: Needs more extensive documentation on how to use this per other components

shared/oae/js/jquery-plugins/jquery.switchtab.js
```diff
@@ -0,0 +1,164 @@
...
+         * OAE plugin that implements the interaction of the "Switch Tab" component. A Switch Tab
+         * is essentially a set of tabs that look like a left-to-right switch. There can only be
+         * one selected at one time. Back functionality is also provided.
+         *
+         * Usage:
+         *
+         *  You should create a template that looks like the following:
+         *
+         *  ```
+         *  <div id="search-scope" class="oae-switchtab-container">
```
nicolaasmatthijs: Rather than rolling our own tab switch component entirely, it would be good if we could leverage the existing Bootstrap pill tabs as much as possible. It is already documented, is semantically a bit better and implements a number of accessibility considerations. 

The `data-switchtab-path` part can probably just be replaced by using the native `href` on the standard tabs component.

We'll probably still want this file and component to bind to and act on some of the events that are sent out by the Bootstrap tabs component.

shared/oae/css/oae.components.css
```diff
@@ -194,6 +194,85 @@ i.fa.large:before {
...
+ ** SWITCH TAB **
+ ****************/
+
+.oae-switchtab-container {
+    text-align: center;
+}
+
+.oae-switchtab-container .oae-switchtab-control {
```
nicolaasmatthijs: If we leverage Bootstrap pill tabs as much as possible, some of these should probably become overrides of their style.

shared/oae/css/oae.components.css
```diff
@@ -194,6 +194,85 @@ i.fa.large:before {
 }
 
 
+/****************
+ ** SWITCH TAB **
+ ****************/
+
+.oae-switchtab-container {
+    text-align: center;
```
nicolaasmatthijs: Could this remain an implementation detail for where the tab switch is being used? If I wanted them to be centered, I would just apply `text-center` to the container I put the tab switch in. If I want them to be left aligned, I wouldn't.

shared/oae/css/oae.components.css
```diff
@@ -194,6 +194,85 @@ i.fa.large:before {
...
+}
+
+.oae-switchtab-container .oae-switchtab-control {
+    background-color: #333;
```
nicolaasmatthijs: All/most of the colors in this component should probably come from the skin file. We could either set up a new section for the switchtab component, or we could leverage some of the existing skin variables:

For example:

- We could use the `oae-tile-metadata` background for the black background. This would then make it align with the tiles
- We could use `.oae-tile > .oae-tile-metadata small` for the text color on the black background. This would then make it align with the tiles
- We could use `.oae-large-options-container label.checked .oae-large-options` for the icon color of the selected tab. This would then make it align with the institutional colors in the visibility options.

I'm not entirely sure which approach I actually prefer. The former approach would be clear and easy to control, but would add another component to the skin and would require us to update all of our existing tenancies. The latter approach would make the institutional colors trickle through to this component without additional work, but could be hard to control.

shared/oae/css/oae.components.css
```diff
@@ -194,6 +194,85 @@ i.fa.large:before {
...
+    margin-right: 35px;
+}
+
+/* Icon generally takes up the most vertical baseline space, so shift it down to help align it with
+   the label content */
+.oae-switchtab-container .oae-switchtab-control .oae-switchtab-switch-icon {
+    top: 1px;
+}
+
+/* Shift the text up one pixel to help align it to larger button. Also give it some space on the
```
nicolaasmatthijs: The latter part of this comment seems to have been removed in code

shared/oae/js/jquery-plugins/jquery.switchtab.js
```diff
@@ -0,0 +1,164 @@
...
+         *
+         * To perform an action when the tab is changed, you should listen to `$(window).on('statechange')`
+         * and use `$mySwitchTab.switchtabId()` to get the tab id of the currently active tab.
+         */
+        $.fn.switchtab = function() {
+            var $switchtab = $(this);
+
+            var currentId = _idFromState($switchtab);
+
+            $switchtab.on('click', '.oae-switchtab-switch', function() {
```
nicolaasmatthijs: I think the bootstrap tabs component will do most of this for you

shared/oae/js/jquery-plugins/jquery.switchtab.js
```diff
@@ -0,0 +1,164 @@
...
+
+            if (!init) {
+                History.pushState(data, title, path);
+            } else {
+                History.replaceState(data, title, path);
+            }
+        };
+
+        /*!
+         * Given an id, set the CSS stat such that the active tab will be labelled as such for
```
nicolaasmatthijs: typo

ui/css/oae.search.css
```diff
@@ -13,36 +13,63 @@
...
 
 #search-form #search-search-icon {
     margin-left: -27px;
     margin-top: 5px;
-    padding: 0px;
     opacity: 0.6;
+
+    /* The typical search icon gets chopped a bit on the left side by the button container, shift it
+       over a little bit */
+    padding: 0 0 0 1px;
```
nicolaasmatthijs: The search icon is too much to the right compared to the icon in the topnavigation. They should be identical

ui/css/oae.search.css
```diff
@@ -13,36 +13,63 @@
...
+}
+
 /* Search container */
 
 #search-content-container {
     padding: 5px 10px;
 }
+
+#search-content-container.search-my .search-refine-type-user,
+#search-content-container.search-my #search-refine-tenant {
```
nicolaasmatthijs: Is there a reason why one is by class and one is by id?

ui/css/oae.search.css
```diff
@@ -13,36 +13,63 @@
...
+
+/* Anonymous only ever has "Everywhere", so don't show the switches at all */
+.anon #search-scope {
+    display: none;
+}
+
 /* Search container */
 
 #search-content-container {
     padding: 5px 10px;
```
nicolaasmatthijs: This should become `15px 10px` in all cases to free up some space at the top

ui/css/oae.search.css
```diff
@@ -13,36 +13,63 @@
...
+
+#search-content-container.search-my .search-refine-type-user,
+#search-content-container.search-my #search-refine-tenant {
+    display: none;
+}
+
+/* Search left-hand navigation */
+
+#search-lhnavigation li > label > input {
+    margin-top: 5px;
```
nicolaasmatthijs: `4px` is slightly better

ui/search.html
nicolaasmatthijs: I'd still be in favour of using the "Home" icon instead of the "Lock" icon, as it might otherwise communicate that "My stuff" will be limited to private items and using the "Home" icon connects it better to "My Personal Space". The "Globe" icon for "Search everywhere" is not ideal in that regard either, but we don't really have a better one.

+1 to adding color to the element icon instead of using the check icon.

ui/search.html
```diff
@@ -23,79 +23,111 @@
         <div data-widget="topnavigation"><!-- --></div>
```
nicolaasmatthijs: When the top navigation is used on the search page, the search bar in the top navigation should be hidden in favour of just the search icon button.

ui/js/search.js
```diff
@@ -13,12 +13,54 @@
...
+     *
+     * @return {Object}     queryData           The query data from the current state
+     * @return {String}     queryData.q         The full-text search query
+     * @return {String[]}   queryData.types     The resource types for which to search
+     * @return {String}     queryData.tenant    The tenant alias in which to search
+     */
+    var _queryData = function() {
+        var params = $.url(History.getState().cleanUrl).param();
+        var q = params.q;
+        var types = [];
```
nicolaasmatthijs: You can swap line 36 and 37

ui/js/search.js
```diff
@@ -13,12 +13,54 @@
...
+
+    /**
+     * Get the search query data in the current state
+     *
+     * @return {Object}     queryData           The query data from the current state
+     * @return {String}     queryData.q         The full-text search query
+     * @return {String[]}   queryData.types     The resource types for which to search
+     * @return {String}     queryData.tenant    The tenant alias in which to search
+     */
+    var _queryData = function() {
```
nicolaasmatthijs: We don't really do `_` for private functions in the UI as pretty much everything is private to the current file/widget

ui/js/search.js
```diff
@@ -28,40 +70,70 @@ require(['jquery','oae.core', 'jquery.history'], function($, oae) {
             infinityScroll.kill();
         }
 
-        // Get the current search query from the History.js data object
-        var query = History.getState().data.query;
-        $('.search-query').val(query);
+        // Ensure the correct tab class is on the search content container
```
nicolaasmatthijs: Might want to clarify what this will be used for

ui/js/search.js
```diff
@@ -28,40 +70,70 @@ require(['jquery','oae.core', 'jquery.history'], function($, oae) {
...
+        // Set the browser title
+        oae.api.util.setBrowserTitle(title);
 
         // Reset the type checkboxes to make sure that none of them stay checked incorrectly
         // when hitting the back and forward buttons
-        $('#search-refine-type input[type="checkbox"]').prop('checked', false);
-        // Get the current type refinements from the History.js data object and select the corresponding checkboxes
-        var types = History.getState().data.types;
-        $.each(types, function(index, type) {
+        $('#search-refine-type input[type="checkbox"]').each(function() {
```
nicolaasmatthijs: No need to do an `each` on this. You can do this as a one-liner

ui/js/search.js
```diff
@@ -72,26 +144,37 @@ require(['jquery','oae.core', 'jquery.history'], function($, oae) {
      * checked/unchecked.
      */
     var modifySearch = function() {
+
```
nicolaasmatthijs: Can remove this line

ui/js/search.js
```diff
@@ -100,25 +183,48 @@ require(['jquery','oae.core', 'jquery.history'], function($, oae) {
...
+        var queryString = url.attr('query');
+
+        /*!
+         * This is backward compatibility handling for links that are sent to OAE. Previously, a
+         * link could be sent to /search/<query> to send a user to the search page with a pre-
+         * loaded query. Now those search endpoints look like /search/all/<query> and
+         * /search/my/<query>.
+         *
+         * To transition to this, we look if the query looks like /search/<query> where <query> is
+         * not "all" or "my". If so, we rewrite the URL to /search/all/<query>.
```
nicolaasmatthijs: Can you add a deprecated notice here and a TODO for removing this in the next release? Can you also send a note to this list about this backwards incompatibility and when it will be removed?

ui/js/search.js
```diff
@@ -127,12 +233,15 @@ require(['jquery','oae.core', 'jquery.history'], function($, oae) {
     var addBinding = function() {
         // Listen to the form submit event on the search form
         $(document).on('submit', '#search-form', modifySearch);
-        // Listen to the change event on the type refinement checkboxes
-        $('#search-refine-type').on('change', 'input[type="checkbox"]', modifySearch);
+
+        // Listen to changes to the checkboxes that refine search options
+        $(document).on('change', '#search-lhnavigation input[type="checkbox"]', modifySearch);
```
nicolaasmatthijs: Why do we need to increase the scope of the listener here?

ui/search.html
```diff
@@ -23,79 +23,111 @@
...
-                                    </button>
-                                </div>
-                            </form>
-                        </div>
+        <main role="main">
+
+            <div class="oae-main-action">
+                <form action="." id="search-form" class="form-search form-horizontal" role="search" aria-label="__MSG__SEARCH__">
+                    <div class="form-group">
+                        <input type="text" id="search-query" class="form-control search-query form-control pull-left" title="__MSG__SEARCH_FOR_CONTENT_FOLDERS_PEOPLE_GROUPS__" autofocus />
```
nicolaasmatthijs: Double `form-control`

ui/search.html
```diff
@@ -23,79 +23,111 @@
...
-                                    <button type="submit" id="search-search-icon" class="btn btn-link pull-left" title="__MSG__SEARCH_FOR_CONTENT_FOLDERS_PEOPLE_GROUPS__">
-                                        <i class="fa fa-search"><span class="sr-only">__MSG__SEARCH__</span></i>
-                                    </button>
-                                </div>
-                            </form>
-                        </div>
+        <main role="main">
+
+            <div class="oae-main-action">
+                <form action="." id="search-form" class="form-search form-horizontal" role="search" aria-label="__MSG__SEARCH__">
```
nicolaasmatthijs: I think `action="."` can be removed?

ui/search.html
```diff
@@ -23,79 +23,111 @@
...
-                            <li>
-                                <label class="checkbox"><input type="checkbox" data-type="discussion" />__MSG__DISCUSSIONS__</label>
-                            </li>
-                        </ul>
+            <div class="oae-main-content">
+                <!-- EVERYWHERE / MY STUFF ONLY -->
+                <div id="search-scope" class="oae-switchtab-container">
+                    <div class="oae-switchtab-control">
+                        <button class="oae-switchtab-switch" data-switchtab-id="all" data-switchtab-path="/search/all">
+                            <span class="oae-switchtab-switch-icon fa fa-globe"></span>
```
nicolaasmatthijs: The icon should be in an `i` tag

ui/search.html
```diff
@@ -23,79 +23,111 @@
...
+                                <li class="search-refine-type-discussion">
+                                    <label class="checkbox">
+                                        <input type="checkbox" data-type="discussion" />
+                                        <span>__MSG__DISCUSSIONS__</span>
+                                    </label>
+                                </li>
+                            </ul>
+                            <ul id="search-refine-tenant" class="nav nav-list">
+                                <li class="nav-header">__MSG__SEARCH_ONLY_WITHIN_COLON__</li>
+                                <li class="search-refine-tenant-${tenant.alias|encodeForHTMLAttribute}">
```
nicolaasmatthijs: We probably don't need the `search-refine-tenant-${tenant.alias|encodeForHTMLAttribute}"` class here?

nicolaasmatthijs:
Finished review. Back to @mrvisser for follow-up.

ui/search.html
sathomas: Actually not. There has to be an `action` attribute for iOS to use the Search keyboard instead of the generic one. (That's what gives you the "Search" key instead of the generic "Go" key.)

ui/js/search.js
sathomas: Yep. I think that will break the STEM incubator.

shared/oae/js/jquery-plugins/jquery.switchtab.js
mrvisser: I don't think using bootstrap tabs JS interaction is the right way to go as they have disabled url changing / back support. I've implemented their HTML template and overridden their styles, but I think that's about as far as we can go.

ui/css/oae.search.css
mrvisser: Yea I believe the containers were given ids (i.e., `#search-refine-tenant` and `#search-refine-type`) and the individual `li` options were given classes (i.e., `.search-refine-tenant-{tenantAlias}` and `.search-refine-type-{resourceType}`.

I probably did this because I didn't feel right putting an ID on the DOM from a dynamic value (tenantAlias and resourceType). If you aren't passionate either way, I'd prefer to just keep it the way it is. Otherwise I'll just change it.

ui/search.html
mrvisser: You had asked me to not touch the topnavigation as you had significant changes coming through. So I just wanted to make sure that you want me to go ahead and do that.

nicolaasmatthijs: Yeah, you're right. Please ignore my comment.

mrvisser: I managed to get something to work by putting a couple styles in `oae.search.css` with an `!important` flag... :)

I'll take it out.

shared/oae/css/oae.components.css
mrvisser: I opted not to add them here in favour of having them in the switchtab component, which I think is the place someone would expect to find docs for that component. The doc format there is similar to the ones here, and I don't think we should have them in both places.

mrvisser:
Followed up and sent a message to list about the deprecation of `/search/<query>`. Back to @nicolaasmatthijs

shared/oae/js/jquery-plugins/jquery.switchtab.js
```diff
@@ -0,0 +1,166 @@
...
+         * OAE plugin that implements the interaction of the "Switch Tab" component. A Switch Tab
+         * is essentially a set of tabs that look like a left-to-right switch. There can only be
+         * one selected at one time. Back functionality is also provided.
+         *
+         * Usage:
+         *
+         * You should create a template that is based on the bootstrap pill component, with the
+         * one exception that the root element has the `oae-switchtab` class:
+         *
+         *  ```
```
nicolaasmatthijs: From previous review:

nicolaasmatthijs:
```
Needs more extensive documentation in oae.component.css on how to use this per other components
```

mrvisser:
```
I opted not to add them here in favour of having them in the switchtab component, which I think is the place someone would expect to find docs for that component. The doc format there is similar to the ones here, and I don't think we should have them in both places.
```

nicolaasmatthijs: I would still prefer this documentation to be added to `oae.components.css` instead. `oae.components.css` is where you can find the documentation for all other component, and the goal is to automatically parse `oae.components.css` and turn it into a widget development component guide at some point

nicolaasmatthijs:
I noticed that there is a regression around the social media authentication buttons. The color of the logo is now grey instead of white.

nicolaasmatthijs:

mrvisser:

nicolaasmatthijs:
Finished another round of review. Back to @mrvisser for follow-up

mrvisser:

mrvisser:
Needs auto-focus on search box on search page

nicolaasmatthijs:
Looks good. I'll create a follow up ticket for the auto-focus.

PR #4014, opened 2015-04-27
Update fr_FR.properties for oae version 11

dooremont:


mrvisser:
@nicolaasmatthijs I'll defer to you but I don't think we can accept this PR, as opposed to syncing from crowdin for a 11.0.1 release.

nicolaasmatthijs:
I've just added the latest i18n keys to Crowdin. My preference is to bring in the translations through a Crowdin synchronisation and there are also a number of additional i18n keys that have been added. Therefore, I'm going to close this PR in favour Crowdin.

PR #4015, opened 2015-05-13
(UserDelete) Add ability to delete a user in user management

mrvisser:
UI Counterpart to https://github.com/oaeproject/3akai-ux/pull/4015


node_modules/oae-admin/manageuser/js/manageuser.js
```diff
@@ -156,15 +160,86 @@ define(['jquery', 'oae.core'], function($, oae) {
...
+                    // Show an error notification
+                    oae.api.util.notification(
+                        oae.api.i18n.translate('__MSG__USER_NOT_DELETED__', 'manageuser'),
+                        oae.api.i18n.translate('__MSG__COULD_NOT_DELETE_USER_NAME__', 'manageuser', {
+                            'displayName': oae.api.util.security().encodeForHTML(widgetData.userProfile.displayName)
+                        }),
+                        'error');
+                } else {
+                    // Success, reload the page
+                    document.location.reload()
```
mrvisser: We *could* be a bit more fancy here and do a live search reload. We would need to close the manageuser modal and then redo the search.

Also, there is a possibility that the search index hasn't been updated by the time all this happens, not exactly sure what we can do about that.

node_modules/oae-admin/manageuser/bundles/default.properties
```diff
@@ -1,9 +1,13 @@
 ACTIONS = Actions
 ADMINISTRATION = Administration
 ADMINISTRATOR = Administrator
+ARE_YOU_SURE_YOU_WANT_TO_DELETE_USER_NAME = Are you sure you want to delete <strong>${displayName}?</strong>
```
nicolaasmatthijs: This can be changed to "Are you sure you want to delete this account?"

node_modules/oae-admin/manageuser/bundles/default.properties
```diff
@@ -1,9 +1,13 @@
...
 BECOME_USER = Become user
 BECOME_USER_NAME = Become ${displayName}
 COULD_NOT_BECOME_USER_NAME = Could not become ${displayName}.
+COULD_NOT_DELETE_USER_NAME = Could not delete ${displayName}.
+DELETE_USER = Delete user
```
nicolaasmatthijs: `DELETE_ACCOUNT = Delete account`

node_modules/oae-admin/manageuser/bundles/default.properties
```diff
@@ -1,9 +1,13 @@
 ACTIONS = Actions
 ADMINISTRATION = Administration
 ADMINISTRATOR = Administrator
+ARE_YOU_SURE_YOU_WANT_TO_DELETE_USER_NAME = Are you sure you want to delete <strong>${displayName}?</strong>
 BECOME_USER = Become user
 BECOME_USER_NAME = Become ${displayName}
 COULD_NOT_BECOME_USER_NAME = Could not become ${displayName}.
+COULD_NOT_DELETE_USER_NAME = Could not delete ${displayName}.
```
nicolaasmatthijs: In order to follow the UI patterns, these messages should be:

```
ACCOUNT_DELETED = Account deleted.
ACCOUNT_DELETE_FAIL = The account could not be deleted.
ACCOUNT_DELETE_SUCCESS = The account has been successfully deleted.
ACCOUNT_NOT_DELETED = Account not deleted.
```

node_modules/oae-admin/manageuser/css/manageuser.css
```diff
@@ -57,8 +57,13 @@
 
 /* ACTIONS */
 
-#manageuser-modal #manageuser-become-user {
-    margin-bottom: 5px;
+#manageuser-modal .checkbox {
```
nicolaasmatthijs: This should be bellow the `#manageuser-become-user` and `#manageuser-delete-user` rules as the checkbox appears bellow them in the UI. We try to maintain the same flow in the CSS as in the UI itself

node_modules/oae-admin/manageuser/js/manageuser.js
nicolaasmatthijs: Infinite scroll has a `removeItems` function that allows you to remove items from an infinite scroll list by id. Rather than reloading the page, I think using that and hiding the modal would be a better experience.

node_modules/oae-admin/manageuser/js/manageuser.js
```diff
@@ -156,15 +160,86 @@ define(['jquery', 'oae.core'], function($, oae) {
...
+            showActionsPanel('main');
+
+            var data = {
                 'context': widgetData.context,
                 'userProfile': widgetData.userProfile
-            }, $('#manageuser-actions-form .modal-body', $rootel));
+            };
+
+            // Render the main panel body
+            oae.api.util.template().render($('#manageuser-actions-body-main-template', $rootel), data,
```
nicolaasmatthijs: We have a pattern that we name a template 'xxx-template' and name the element in which the rendered template will be added 'xxx-container'

node_modules/oae-admin/manageuser/js/manageuser.js
```diff
@@ -156,15 +160,86 @@ define(['jquery', 'oae.core'], function($, oae) {
...
+                $('#manageuser-actions-form .modal-body .manageuser-actions-main', $rootel));
+
+            // Render the deleteuser panel body and footer
+            oae.api.util.template().render($('#manageuser-actions-body-deleteuser-template', $rootel), data,
+                $('#manageuser-actions-form .modal-body .manageuser-actions-deleteuser', $rootel));
+
+            oae.api.util.template().render($('#manageuser-actions-footer-deleteuser-template', $rootel), data,
```
nicolaasmatthijs: Given the design changes, this will no longer have to be a template

node_modules/oae-admin/manageuser/js/manageuser.js
```diff
@@ -156,15 +160,86 @@ define(['jquery', 'oae.core'], function($, oae) {
...
+                $('#manageuser-actions-form .modal-footer .manageuser-actions-deleteuser', $rootel));
+
+            // Switch to deleteuser panel when user selects to delete the user
+            $rootel.on('click', '#manageuser-delete-user', function() {
+                showActionsPanel('deleteuser');
+            });
+
+            // Delete the user when the user accepts the deleteuser confirmation
+            $rootel.on('click', '#manageuser-actions-form .manageuser-deleteuser-confirm', deleteUser);
```
nicolaasmatthijs: This should go in `addBinding`

nicolaasmatthijs: I would prefer the confirm button and cancel button to have ids, as they will only be in the page once. In that case, there should also be no need to prefix with `#manageuser-actions-form`

node_modules/oae-admin/manageuser/manageuser.html
```diff
@@ -46,10 +46,16 @@ <h3 id="manageuser-modal-title"><!-- --></h3>
                 </form>
 
                 <form id="manageuser-actions-form" class="tab-pane form-horizontal" role="form">
```
nicolaasmatthijs: I think the following modal structure would be better:

```
<form id="manageuser-actions-form" class="tab-pane form-horizontal" role="form">
  <div id="manageuser-actions-main">
    <div class="modal-body">
      ...
    </div>
    <div class="modal-footer">
      ...
    </div>
  </div>
  <div id="manageuser-actions-deleteuser">
    <div class="modal-body">
      ...
    </div>
    <div class="modal-footer">
      ...
    </div>
  </div>
</form>
```

Hiding and showing becomes a matter of switching the visible div, and they can be cleanly identified by id

node_modules/oae-admin/manageuser/manageuser.html
```diff
@@ -199,6 +209,16 @@ <h3 id="manageuser-modal-title"><!-- --></h3>
     </div>
 --></div>
 
+<div id="manageuser-actions-body-deleteuser-template"><!--
```
nicolaasmatthijs: This would be the desired display for the delete user confirmation dialog:

screen shot 2015-05-19 at 13 33 55

node_modules/oae-admin/manageuser/manageuser.html
nicolaasmatthijs: You could even put them in 2 forms and hide/show the forms. The advantage is that you can then bind to the submit action of both forms, preventing to have bind to the button click.

node_modules/oae-admin/manageuser/js/manageuser.js
nicolaasmatthijs: After the suggested changes, it should no longer be necesasry to reference the submit button

shared/oae/api/oae.api.admin.js
```diff
@@ -292,6 +292,34 @@ define(['exports', 'jquery', 'underscore'], function(exports, $, _) {
     };
 
     /**
+     * Delete a user
```
nicolaasmatthijs: Should we add a note about the user being soft deleted and the ability to restore the user?

nicolaasmatthijs:

node_modules/oae-admin/manageuser/js/manageuser.js
mrvisser: That seems like a better experience and more robust.

node_modules/oae-admin/manageuser/bundles/default.properties
mrvisser: Should the become user stuff change to a similar format as well to follow the UI patterns?

`USER_BECAME`, `USER_BECOME_FAIL` etc...?

node_modules/oae-admin/manageuser/js/manageuser.js
mrvisser: It doesn't appear anything in this `manageuser` widget actually follows that. I'll just change all the ids to save time in the future.

mrvisser:

nicolaasmatthijs:
Finished another round of review. Back to @mrvisser for review.

mrvisser:

nicolaasmatthijs:

mrvisser:

nicolaasmatthijs:
Finished review. Back to @mrvisser for final outstanding comments

mrvisser:

nicolaasmatthijs:
Merged.

PR #3701, opened 2014-04-01
(Issue 3453) Fix remove button not working on overflowing autosuggestion item

bp323:

PR #4017, opened 2015-05-20
Fix merge conflict in https://github.com/oaeproject/3akai-ux/pull/3701

stuartf:


stuartf:
This is just a fix of the merge conflict in #3701

shared/oae/css/oae.base.css
```diff
@@ -464,14 +464,24 @@ ul.as-selections li.as-original input::-ms-clear {
...
+            border-radius: 3px;
+    -webkit-box-shadow: none;
+       -moz-box-shadow: none;
+            box-shadow: none;
     font-family: inherit;
     line-height: 30px;
     margin-bottom: 3px;
     margin-right: 5px;
-    max-width: 400px;
+    max-width: 300px;
```
nicolaasmatthijs: If the span has a `max-width` and a `width: auto`, I don't think the `li` needs them

shared/oae/css/oae.base.css
```diff
@@ -464,14 +464,24 @@ ul.as-selections li.as-original input::-ms-clear {
 }
 
 ul.as-selections li.as-selection-item {
-    border-radius: 3px;
-    box-shadow: none;
+    -webkit-border-radius: 3px;
```
nicolaasmatthijs: We don't do vendor prefixes for `border-radius`, so these can be removed

shared/oae/css/oae.base.css
```diff
@@ -464,14 +464,24 @@ ul.as-selections li.as-original input::-ms-clear {
...
+       -moz-border-radius: 3px;
+            border-radius: 3px;
+    -webkit-box-shadow: none;
```
nicolaasmatthijs: I don't think we need to do vendor prefixes for `box-shadow` either. You can remove these + any other `box-shadow` vendor prefixes we have

nicolaasmatthijs:
Finished review. Back to @stuartf for follow-up

stuartf:
fixed review issues

nicolaasmatthijs:
Looks good. Merged.

PR #4018, opened 2015-05-27
Use our own URL expander and update the jQuery oembed plugin

simong:
UI counter part of https://github.com/oaeproject/Hilary/pull/1139


simong:
I'm marking this as `On hold` as it's currently relying on a PR against the jquery-oembed-all repository: https://github.com/nfl/jquery-oembed-all/pull/22

mrvisser:
@simong things seem to be moving a bit slow on their end. Did you want to switch to a github clone for now?

simong:
I'm OK with that. My copy of the changes are already included in this PR

mrvisser:
Ah right, this isn't NPM.

mrvisser:
@simong can you add a note in the header of the oembed file that it is forked, with a reference to the external PR that is pending review?

PR #4021, opened 2015-05-28
When an error occurs during manageacces, removed members still stay cached for adding users

sathomas:
1. Create a group
2. Click manage access
3. Add several users, including the group itself
    * Note that another issue we should resolve is that the group itself doesn't show up in auto-suggest
    * This serves as a way to trigger a 400 error when you save the membership changes
4. Save the membership changes, there is an error because the group was being added to itself
5. Remove the group from the list
6. Click save. Still an error because it's still trying to add the group
7. Remove all the other users you tried to add and click save, all the members you removed are still part of the POST request

### DIFF
--- a/node_modules/oae-core/manageaccess/js/manageaccess.js
+++ b/node_modules/oae-core/manageaccess/js/manageaccess.js
@@ -27,10 +27,24 @@ define(['jquery', 'oae.core', 'jquery.autosuggest'], function($, oae) {
         // Caches the visibility of the context (e.g. group/content/discussion/...)
         var visibility = null;
 
-        // Caches the member updates that need to be applied
+        // Caches the member updates that need to be applied. Each key is a
+        // member (user or group) id and its value is the role for that member.
+        // A value of `false` deletes the member from the group entirely. This
+        // object is sent to the server via an API call, so all updates should
+        // be relative to the server state.
         var membersUpdates = {};
         
-        // Caches members pending addition
+        // Caches members pending addition. This is distinct from `membersUpdates`
+        // to handle deletion. If a member that the user decides to delete is
+        // already a full member of the group, then we have to use the API to
+        // delete that member on the server. If, on the other hand, a member
+        // that the user decides to delete is one that the user has just added
+        // (but not saved to the server), then we don't use the API. Each key
+        // is a member (user or group) id that the user has identified as a
+        // candidate for addition; however, the member has not beend added to
+        // the group (yet) on the server. The value is normally `true` but can
+        // be set to `false` if the user changes his mind and removes the member
+        // before the server is updated.
         var pendingMembers = {};
 
         // Variable that will be used to keep track of the current infinite scroll instance
--- a/node_modules/oae-core/manageaccess/js/manageaccess.js
+++ b/node_modules/oae-core/manageaccess/js/manageaccess.js
@@ -220,7 +220,8 @@ define(['jquery', 'oae.core', 'jquery.autosuggest'], function($, oae) {
          */
         var setUpAutoSuggest = function() {
             oae.api.util.autoSuggest().setup($('#manageaccess-share-autosuggest', $rootel), {
-                'selectionChanged': autoSuggestChanged
+                'selectionChanged': autoSuggestChanged,
+                'exclude': widgetData.contextProfile.id
             }, null, function() {
                 // Focus on the autosuggest field once it has been set up
                 focusAutoSuggest();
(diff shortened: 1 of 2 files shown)


sathomas:
over to @mrvisser for review

node_modules/oae-core/manageaccess/js/manageaccess.js
```diff
@@ -175,7 +179,12 @@ define(['jquery', 'oae.core', 'jquery.autosuggest'], function($, oae) {
             infinityScroll.removeItems(principalId);
 
             // Mark the member as deleted
-            membersUpdates[principalId] = false;
+            if (pendingMembers[principalId]) {
+                delete membersUpdates[principalId];
+                pendingMembers[principalId] = false;
+            } else {
+                membersUpdates[principalId] = false;
+            }
```
mrvisser: If we get rid of `pendingMembers` and do:

```javascript
if(membersUpdates[principalId]) {
    delete membersUpdates[principalId];
} else {
    membersUpdates[principalId] = false;
}
```

Would that work the same?

mrvisser: No I don't think it will work the same because if we updated a role of an existing user then decided to remove them after, it wouldn't perform the remove.

shared/oae/api/oae.api.util.js
```diff
@@ -960,10 +968,24 @@ define(['exports', 'require', 'jquery', 'underscore', 'oae.api.config', 'markdow
...
                     // for finding out the search query
                     var query = $.url(this.url).param('q');
+                    
+                    // Track any results that need to be excluded
+                    var toExclude = [];
+
                     $.each(data.results, function(index, result) {
-                        result.displayName = security().encodeForHTML(result.displayName);
-                        result.query = query;
+                        if (_.indexOf(options.exclude, result.id) !== -1) {
```
mrvisser: I think `!_.contains(options.exclude, result.id)` is more representative of the check

shared/oae/api/oae.api.util.js
```diff
@@ -960,10 +968,24 @@ define(['exports', 'require', 'jquery', 'underscore', 'oae.api.config', 'markdow
...
+                            toExclude.push(index);
+                        } else {
+                            result.displayName = security().encodeForHTML(result.displayName);
+                            result.query = query;
+                        }
                     });
+                    
+                    // Remove excluded items
+                    while (toExclude.length > 0) {
```
mrvisser: `while (!_.isEmpty(toExclude)) {`

We're pretty high on underscore. Maybe even moreso than when you last did direct OAE stuff :)

shared/oae/api/oae.api.util.js
```diff
@@ -960,10 +968,24 @@ define(['exports', 'require', 'jquery', 'underscore', 'oae.api.config', 'markdow
...
+                        data.results.splice(_.last(toExclude), 1);
+                        toExclude.pop();
```
mrvisser: We can get rid of a line if you just do `data.results.splice(toExclude.pop(), 1);`. But understand if you don't want to go that route since it may conceal the mutation of the array.

node_modules/oae-core/manageaccess/js/manageaccess.js
sathomas: Right. I can't think of a clean way to combine the objects (without changing the API utility code).

node_modules/oae-core/manageaccess/js/manageaccess.js
```diff
@@ -29,6 +29,9 @@ define(['jquery', 'oae.core', 'jquery.autosuggest'], function($, oae) {
 
         // Caches the member updates that need to be applied
         var membersUpdates = {};
+        
+        // Caches members pending addition
+        var pendingMembers = {};
```
mrvisser: Can you make this an array and call it something like `newUserIds`? Also update the docs on `membersUpdates` to better indicate that the values of the object will be the role updates to apply. That way the purpose of the two objects are better distinguished

mrvisser: I guess making it an array isn't entirely necessary since you want them to be unique. But reading through the difference of these isn't entirely clear. Maybe just needs more descriptive docs.

shared/oae/api/oae.api.util.js
sathomas: Actually, I was trying to highlight that you have to process excluded items from the end of the array to the beginning to preserve the indices, but I can do that better with a comment.

node_modules/oae-core/manageaccess/js/manageaccess.js
sathomas: Sure. I'll update the docs. I originally implemented it as an array, but handling all the possible adds and removes got a little messy.

PR #4023, opened 2015-06-01
Extend userdetails to catch display names with @

sathomas:
The userdetails widget should be extended to also consider display names with an `@` to be invalid. When going through the Shibboleth responses that we've received so far, I noticed quite a few that only release a targeted-id in the form of `...=@surrey.ac.uk`.

### DIFF
--- a/shared/oae/api/oae.api.util.js
+++ b/shared/oae/api/oae.api.util.js
@@ -589,8 +589,11 @@ define(['exports', 'require', 'jquery', 'underscore', 'oae.api.config', 'markdow
             if (options.submitHandler && $.isFunction(options.submitHandler)) {
                 submitCallback = options.submitHandler;
             }
+            
+            // Register common custom validation methods
+            $.validator.addMethod('displayname', isValidDisplayName, require('oae.api.i18n').translate('__MSG__NAMES_CANNOT_INCLUDE_URLS_OR_AT_SIGNS__'));
 
-            // Register the custom validation methods
+            // Register additional custom validation methods
             if (options.methods) {
                 $.each(options.methods, function(key, value) {
                     $.validator.addMethod(key, value.method, value.text);
--- a/shared/oae/api/oae.api.util.js
+++ b/shared/oae/api/oae.api.util.js
@@ -701,6 +701,11 @@ define(['exports', 'require', 'jquery', 'underscore', 'oae.api.config', 'markdow
             if (/https?:\/\//i.test(displayName)) {
                 return false;
             }
+            
+            // Display names that contain `@` are also considered invalid
+            if (/@/.test(displayName)) {
+                return false;
+            }
 
             return true;
         };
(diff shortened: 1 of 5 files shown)


sathomas:
To @nicolaasmatthijs for review. I've assumed that if we want to prevent a user from manually entering an `@` character in the form, that we can rely on server validation via `oae.api.user.updateUser` (as with `http://` currently).

shared/oae/api/oae.api.util.js
```diff
@@ -701,6 +701,11 @@ define(['exports', 'require', 'jquery', 'underscore', 'oae.api.config', 'markdow
             if (/https?:\/\//i.test(displayName)) {
                 return false;
             }
+            
+            // Display names that contain `@` are also considered invalid
```
nicolaasmatthijs: Display names containing an `@` sign are also indicative of Shibboleth not releasing an attribute that could be used as the display name. Can we change this to:

```
// Display names that contain `http://`, `https://` or `@` are indicative of Shibboleth
// not releasing an attribute that could be used as the display name. This is not
// considered to be valid
if (/https?:\/\//i.test(displayName)) {
    return false;
} else if (/@/.test(displayName)) {
    return false;
}
```

shared/oae/api/oae.api.util.js
```diff
@@ -701,6 +701,11 @@ define(['exports', 'require', 'jquery', 'underscore', 'oae.api.config', 'markdow
             if (/https?:\/\//i.test(displayName)) {
```
nicolaasmatthijs: I noticed that the form validation in the `userdetails` widget does not appear to be using `isValidDisplayName` for validation of the entered display name. Because of that, it's now possible to add names containing `http(s)` or `@`.

nicolaasmatthijs:

sathomas:
followed up on review. back to @nicolaasmatthijs

shared/oae/bundles/ui/default.properties
```diff
@@ -398,6 +398,7 @@ MY_PICTURE = My picture
 MY_PREFERENCES = My preferences
 MY_PROFILE = My profile
 NAME = Name
+NAMES_CANNOT_INCLUDE_URLS_OR_AT_SIGNS = Names cannot include URLs or the '@' character
```
nicolaasmatthijs: Can you change the message to "Please enter a valid name" ?

shared/oae/api/oae.api.util.js
```diff
@@ -589,8 +589,11 @@ define(['exports', 'require', 'jquery', 'underscore', 'oae.api.config', 'markdow
             if (options.submitHandler && $.isFunction(options.submitHandler)) {
                 submitCallback = options.submitHandler;
             }
+            
+            // Register common custom validation methods
+            $.validator.addMethod('displayname', isValidDisplayName, require('oae.api.i18n').translate('__MSG__NAMES_CANNOT_INCLUDE_URLS_OR_AT_SIGNS__'));
```
nicolaasmatthijs: Rather than making this validator globally available, I'd prefer to add it as a custom validator inside of the `userdetails` widget. The `register` widget has some good examples on how to do this as part of `setUpValidation`

nicolaasmatthijs:

sathomas:
name validation more generic; back to @nicolaasmatthijs

nicolaasmatthijs:
Looks good. Merged.

PR #4027, opened 2015-06-11
List item improvements

sathomas:
- Rather than only wrapping a link around the thumbnail and the title of the item, we should wrap the entire tile in a link (this should not be he case for the other list types though)
- In the User Management page in the admin UI, clicking the bottom of a tile will actually cause you to click the list item below. This for example makes it difficult to click the `Import CSV` button

screen shot 2015-06-11 at 07 15 55

### DIFF
--- a/shared/oae/css/oae.components.css
+++ b/shared/oae/css/oae.components.css
@@ -387,9 +387,7 @@ div.oae-thumbnail {
 
 div.oae-thumbnail:before {
     display: inline-block;
-    font-size: 18px;
-    margin-top: -9px;
-    padding-top: 50%;
+    position: relative;
 }
 
 div.oae-thumbnail a,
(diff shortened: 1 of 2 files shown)


sathomas:
@nicolaasmatthijs: this PR only fixes the second of the two issues, but it's a rather significant change to the list macro, so I wanted to let you check it out. Either I don't understand something (likely), or there were some serious problems lurking in the macro that we haven't addressed until now.

mrvisser:
@sathomas I recall this area being pretty hairy for X-browser compatibility, hence a lot of the very specific pixel numbering. If you haven't already, the impact should be spot-checked on all browsers, particularly IE9 and 10, and all the different list item sizes, as I think the change to the `oae-thumbnail` class will impact all of them.

sathomas:
Yes. It's worth checking, but a couple of things about the old code don't make sense:

1. Why use "small" instead of "medium" for the detailed view?
2. Why doesn't the compact view set any display options?

Also, can we finally drop IE9 support for good? It seems like it's way past time.

sathomas:
And, in any case, the "very specific pixel numbering" is horribly broken, as this issue attests.

PR #4028, opened 2015-06-15
Manageuser widget should provide list of all associated LoginIds for a user

Coenego:
The manageuser widget is an admin UI widget that allows a tenant administrator to view and change details of a particular user. The last tab in that widget is currently the "Password" tab, allowing for the local password of that user to be changed.

This tab needs to be changed to an `Authentication` tab. The content of this tab should be a list of all of the associated LoginIds for that user. For each LoginId, we should show the type (cas, shibboleth, facebook, local) and the external id for that type. If the user has a local LoginId, this should be shown at the bottom of the list accompanied by the "change password" form. If the user does not have a local LoginId, the "change password" form should not be shown.

screen shot 2015-06-11 at 23 21 32

### DIFF
--- a/node_modules/oae-admin/manageuser/manageuser.html
+++ b/node_modules/oae-admin/manageuser/manageuser.html
@@ -25,7 +25,7 @@
                             <a href="#manageuser-actions-form" data-toggle="tab">__MSG__ACTIONS__</a>
                         </li>
                         <li>
-                            <a href="#manageuser-password-form" data-toggle="tab">__MSG__PASSWORD__</a>
+                            <a href="#manageuser-authentication-form" data-toggle="tab">__MSG__AUTHENTICATION__</a>
                         </li>
                     </ul>
                 </div>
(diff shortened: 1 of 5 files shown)


Coenego:
Assigning to @mrvisser for review

node_modules/oae-admin/manageuser/manageuser.html
```diff
@@ -98,6 +83,43 @@ <h3 id="manageuser-editprofile-header-container"><!-- --></h3>
     </div>
 </div>
 
+<div id="manageuser-authentication-template"><!--
+    {if loginIds}
```
mrvisser: Do we need to check this? I feel like we can always have a `loginIds` property available.

Coenego: Good point

mrvisser:
Code-convention-wise this looks good to me, but I think it needs some design feedback.

A couple things that come to mind:

1. If the user only has a non-local authentication id, things look lopsidedly aligned to the right
2. I don't think the external id should sit in a disabled text field
3. I think the containers should somehow be styled to the strategy signing we see in the `Sign in` panel, including organizational logo, if applicable
4. We should probably apply a conscious ordering to the login ids aside from just the local one at the bottom (e.g., always have organizational shib/cas at the top, then maybe alphabetically ordered social strategies, then local)

Assigning to @nicolaasmatthijs for feedback

mrvisser:
A couple screenshots of what is there:

Twitter only:

screen shot 2015-06-15 at 8 08 31 am

Local only:

screen shot 2015-06-15 at 8 08 42 am

sallakarppinen:
I don't think we currently allow for more than one loginId per user, but as that was a part of the original requirements I have tested such a situation with the following result: 
screenshot from 2017-01-25 14-54-48

Obviously, the second `Twitter` field should have the local username instead. Will look into fixing this.

sallakarppinen:
Closed in favour of #4117

PR #3932, opened 2015-06-17
(A11y severity 3) Remove hidden heading from pages

sathomas:
Several pages within the site have a hidden first- level heading titled "Content begins here". Since the pages already have a <main> element with an ARIA role of main, the hidden first-level heading is unnecessary and should be removed.

### DIFF
--- a/node_modules/oae-core/lhnavigation/lhnavigation.html
+++ b/node_modules/oae-core/lhnavigation/lhnavigation.html
@@ -5,7 +5,7 @@
     <div class="oae-lhnavigation">
         <ul class="nav nav-list"><!-- --></ul>
     </div>
-    <div class="oae-page"><!-- --></div>
+    <div id="oae-page-content" class="oae-page"><!-- --></div>
 </div>
 
 <div id="lhnavigation-navigation-template"><!--
--- a/node_modules/oae-core/topnavigation/bundles/default.properties
+++ b/node_modules/oae-core/topnavigation/bundles/default.properties
@@ -1,5 +1,3 @@
-SKIP_TOP_MENU = Skip to content [c]
-SKIP_TOP_MENU_KEY = c
-SKIP_MENU_TARGET = Content begins here
+SKIP_TO_CONTENT = Skip to content
 TOPNAV_ARIA_LABEL = Global navigation bar
 TOPNAV_SEARCH_ARIA_LABEL = Search everything
--- a/node_modules/oae-core/topnavigation/js/topnavigation.js
+++ b/node_modules/oae-core/topnavigation/js/topnavigation.js
@@ -112,6 +112,11 @@ define(['jquery', 'oae.core', 'activityadapter'], function($, oae, ActivityAdapt
          * to jump past the top navigation
          */
         var setUpSkipLinks = function() {
+            // Index page has a unique structure.
+            if ($('#index-page-container').length > 0) {
+                $('.oae-skip-link',$rootel).attr('href', '#index-page-container');
+            }
+
             // Focus on the target element when clicking
             // a skip link
             $rootel.on('click', '.oae-skip-link', function() {
--- a/node_modules/oae-core/topnavigation/topnavigation.html
+++ b/node_modules/oae-core/topnavigation/topnavigation.html
@@ -25,7 +25,6 @@
         </div>
     </div>
 </div>
-<h1 id="topnavigation-target" class="sr-only">__MSG__SKIP_MENU_TARGET__</h1>
 
 <div id="topnavigation-institutional-logo-template"><!--
     <a href="/" class="oae-institutional-logo hidden-xs visible-sm" title="${oae.data.me.tenant.displayName|encodeForHTMLAttribute}">
--- a/node_modules/oae-core/topnavigation/topnavigation.html
+++ b/node_modules/oae-core/topnavigation/topnavigation.html
@@ -3,7 +3,7 @@
 
 <div id="topnavigation-container" role="navigation" aria-label="__MSG__TOPNAV_ARIA_LABEL__">
     <!-- ACCESSIBILITY HELPERS -->
-    <a id="topnavigation-skip" class="sr-only oae-skip-link hide" href="#topnavigation-target" accesskey="__MSG__SKIP_TOP_MENU_KEY__">__MSG__SKIP_TOP_MENU__</a>
+    <a id="topnavigation-skip" class="sr-only oae-skip-link hide" href="#oae-page-content">__MSG__SKIP_TO_CONTENT__</a>
     <!-- NAVIGATION -->
     <div class="row">
         <!-- LEFT MENU -->


sathomas:
ready for @nicolaasmatthijs review. Fixes #3935 as well

node_modules/oae-core/lhnavigation/lhnavigation.html
```diff
@@ -5,7 +5,7 @@
     <div class="oae-lhnavigation">
         <ul class="nav nav-list"><!-- --></ul>
     </div>
-    <div class="oae-page"><!-- --></div>
+    <div id="oae-page-content" class="oae-page"><!-- --></div>
```
nicolaasmatthijs: Can you confirm that jumping to the actual content of the page is the intended behavior? The previous behavior just made it jump until right after the top navigation (but before the left hand navigation).

node_modules/oae-core/topnavigation/bundles/default.properties
```diff
@@ -1,5 +1,3 @@
-SKIP_TOP_MENU = Skip to content [c]
-SKIP_TOP_MENU_KEY = c
```
nicolaasmatthijs: These need to be removed from the other languages as well

node_modules/oae-core/topnavigation/bundles/default.properties
```diff
@@ -1,5 +1,3 @@
...
-SKIP_MENU_TARGET = Content begins here
+SKIP_TO_CONTENT = Skip to content
```
nicolaasmatthijs: Is not having an accesskey preferable over having one?

node_modules/oae-core/topnavigation/js/topnavigation.js
```diff
@@ -112,6 +112,11 @@ define(['jquery', 'oae.core', 'activityadapter'], function($, oae, ActivityAdapt
          * to jump past the top navigation
          */
         var setUpSkipLinks = function() {
+            // Index page has a unique structure.
+            if ($('#index-page-container').length > 0) {
```
nicolaasmatthijs: Rather than referencing the index page in the top navigation widget, can we just add the `oae-page-content` id to the appropriate point in the index page? When doing this, we might want to make that id a bit more generic (e.g. `oae-main-content`)

node_modules/oae-core/topnavigation/js/topnavigation.js
```diff
@@ -112,6 +112,11 @@ define(['jquery', 'oae.core', 'activityadapter'], function($, oae, ActivityAdapt
...
+                $('.oae-skip-link',$rootel).attr('href', '#index-page-container');
+            }
+
             // Focus on the target element when clicking
             // a skip link
             $rootel.on('click', '.oae-skip-link', function() {
```
nicolaasmatthijs: I thought it wasn't possible to put focus on a div?

nicolaasmatthijs:

node_modules/oae-core/lhnavigation/lhnavigation.html
sathomas: If there's only one skip link, then it should go to the main content.

[WebAIM](http://webaim.org/techniques/skipnav/):

> provide a link at the top of the page which jumps the user down to an anchor or target at the beginning of the main content.

[Web Accessibility](https://www.webaccessibility.com/best_practices.php?best_practice_id=543):

> The skip link should be located at the beginning of the page and should target the unique content of the page.

Some sites have two (or more) skip links, e.g. one to skip the header and another to skip to content. The text for our skip links has always been "Skip to content."

node_modules/oae-core/topnavigation/js/topnavigation.js
sathomas: * The target of an `#href` has to be an `id`, and we already have an `id` assigned to the content block on the index page.

* We already have an `oae-main-content` class that's applied to the `<main>` element. It would seem to be confusing to have that class on one element and an `id` with the same name associated with a different element.

node_modules/oae-core/topnavigation/js/topnavigation.js
sathomas: It is possible if the `div` has a `tabindex`

sathomas:
back to @nicolaasmatthijs after followup

node_modules/oae-core/topnavigation/js/topnavigation.js
nicolaasmatthijs: My preference would still be to think of a name different than `oae-main-content` and use that as the id for the main content container in both the lhnav widget and index page. I think that's cleaner than introducing an implicit dependency between several widgets.

node_modules/oae-core/topnavigation/bundles/ca_ES.properties
```diff
@@ -1,6 +1,4 @@
 #X-Generator: crowdin.com
-SKIP_TOP_MENU=Anar a la patgina [c]
-SKIP_TOP_MENU_KEY=c
 SKIP_MENU_TARGET=El contingut comen\u00e7a a\u00e7i
```
nicolaasmatthijs: In the default.properties file, we are retaining the `SKIP_TO_CONTENT` key, whilst in the other language files we are retaining the `SKIP_MENU_TARGET` key

ui/index.html
```diff
@@ -23,7 +23,7 @@
 
         <!-- BODY -->
         <main id="index-container" class="oae-branding-container oae-main-content" role="main">
-            <div id="index-page-container" class="row"><!-- --></div>
+            <div id="index-page-container" tabindex="0" class="row"><!-- --></div>
```
nicolaasmatthijs: Is it common practice to have a div be the target for a skip to content link? Wouldn't it be better to link to an element that would provide some content (e.g. header that says `Content begins here`) to the screenreader user (and would be natively focusable) ?

nicolaasmatthijs:
Finished another round of review. Back to @sathomas

ui/index.html
sathomas: The examples I've found, e.g. [OSU Accessibility](http://oregonstate.edu/accessibility/skiplinks) or [WebAIM Guidelines](http://webaim.org/techniques/skipnav/) just use a standard block element as the skip link target. (They don't add the `tabindex` either but I don't see any harm in it.)

I think that screen readers might find the extra speech ("Content begins here") slightly annoying. OTOH, it would provide confirmation that the skip link worked as expected.

sathomas:
back to @nicolaasmatthijs

node_modules/oae-core/topnavigation/js/topnavigation.js
nicolaasmatthijs: What about something like `oae-skip-content-target` ?

ui/index.html
nicolaasmatthijs: If the examples don't add `tabindex`, I'd be happy removing it as well. The extra tab stop becomes somewhat annoying when using the keyboard with the tabindex in place.

node_modules/oae-core/topnavigation/topnavigation.html
```diff
@@ -3,7 +3,7 @@
 
 <div id="topnavigation-container" role="navigation" aria-label="__MSG__TOPNAV_ARIA_LABEL__">
     <!-- ACCESSIBILITY HELPERS -->
-    <a id="topnavigation-skip" class="sr-only oae-skip-link hide" href="#topnavigation-target" accesskey="__MSG__SKIP_TOP_MENU_KEY__">__MSG__SKIP_TOP_MENU__</a>
+    <a id="topnavigation-skip" class="sr-only oae-skip-link hide" href="#oae-page-content">__MSG__SKIP_TO_CONTENT__</a>
```
nicolaasmatthijs: This link has the `hide` class applied to it, causing it to never receive focus. I'm not sure why it was ever added and if it ever worked.

nicolaasmatthijs:

sathomas:
pushed some changes but I'm not sure it's what @nicolaasmatthijs wants.

There is already an `id` assigned to the index page content and, e.g., all of the styles and some javascript (template rendering) are based on it. It wouldn't make sense to re-use that `id` (`index-page-container`) on other pages, but renaming it something more generic would violate the CSS conventions we've established.

In any case having an implicit dependency between several widgets is, I think, unavoidable. If we change the `id` in the index page, then both the code and the styles for that page will have an implicit dependency on the topnav widget. If we leave the index page alone, then the code in the topnav widget will have an implicit dependency on the index page.

nicolaasmatthijs:
Once a PR for https://github.com/oaeproject/3akai-ux/issues/4040 is available, can you rebase this branch against that branch? I believe that the changes described in #4040 will remove the need for the special index page handling.

sathomas:
On hold pending disposition of #4040

PR #3938, opened 2015-06-17
(A11y severity 3) Remove redundancy from search box

sathomas:
The search box in the top right of all pages has a hidden `<label>`, an `aria-label`, and `title` attribute that all label the field. This causes lots of redundant information to be presented to screen readers (title is read by some screen readers in addition to labels). Remove the redundant `aria-label` and `title`.


sathomas:
assigning to @nicolaasmatthijs for review

node_modules/oae-core/topnavigation/topnavigation.html
```diff
@@ -13,10 +13,10 @@
         <!-- RIGHT MENU -->
         <div id="topnavigation-right" class="col-xs-5 text-right">
             <!-- SEARCH FORM -->
-            <form action="." id="topnavigation-search-form" class="form-inline" role="search" aria-label="__MSG__TOPNAV_SEARCH_ARIA_LABEL__">
```
nicolaasmatthijs: Can you remove the `action="."`?

node_modules/oae-core/topnavigation/topnavigation.html
```diff
@@ -13,10 +13,10 @@
...
+            <form action="." id="topnavigation-search-form" class="form-inline" role="search">
                 <div class="form-group">
                     <label for="topnavigation-search-query" class="control-label sr-only">__MSG__SEARCH__</label>
-                    <input type="text" id="topnavigation-search-query" name="topnavigation-search-query" class="search-query form-control pull-left" title="__MSG__SEARCH_FOR_CONTENT_FOLDERS_PEOPLE_GROUPS__"/>
+                    <input type="text" id="topnavigation-search-query" name="topnavigation-search-query" class="search-query form-control pull-left" />
```
nicolaasmatthijs: Removing the title causes it to no longer have a tooltip. Given that the search field doesn't have a visible label or placeholder, wouldn't it be better to keep the title in place?

nicolaasmatthijs:
Finished review. Back to @sathomas

sathomas:
We could remove the hidden label instead of the `title` attribute, but some screen readers won't read `title` text, so that would make the search inaccessible to those users. Alternatively we could use the Bootstrap [tooltip](http://getbootstrap.com/javascript/#tooltips) component instead of the `title` attribute. @nicolaasmatthijs would you like me to update the PR with a Bootstrap implementation to check out, or would you rather keep it simple and delete the label?

nicolaasmatthijs:
@sathomas : Would you mind reaching out to WebAIM to check if it is acceptable to have both a hidden label and a title attribute for tooltip purposes? I'm assuming removing the hidden label is not going to be acceptable, but it might be worth checking with them.

sathomas:
It turns out that things aren't so clear cut. There's an extended discussion of this problem on the [WebAIM mailing list](http://webaim.org/discussion/mail_thread?thread=6213), but the conclusion seems to be that it's okay to use a `title` instead of a `<label>` if you want the tooltip.

Corroboration from [WebAIM Guidelines](http://webaim.org/techniques/forms/advanced):

> If a form field has a title attribute, but no <label>, the screen reader will read the title as if it were a label.

Further support from [The Paciello Group](http://www.paciellogroup.com/blog/2011/02/html5-accessibility-chops-the-placeholder-attribute/)

> **Always** provide an explicitly associated text label for a text field, using either the `label` element or the `title` attribute.

It seems like the biggest disadvantage to using `title` is that sighted keyboard-only users will never see it. They will, of course, see the magnifying glass icon, but I suppose there might be some cultures in which that iconography isn't clear. If we want to address that concern, we could add a `placeholder` attribute to the field.

I've updated the PR by removing the `<label>` element instead of the `title` attribute. Back to @nicolaasmatthijs for thoughts, review, etc.

nicolaasmatthijs:
Looks good. Merged.

PR #3937, opened 2015-06-18
(A11y severity 3) Close clips on loss of focus

sathomas:
There are a number of buttons that expand when selected to reveal additional options (typically with a class that includes "clip-fold"). Once a user navigates out of these boxes, the sections remain open, blocking content beneath them from view. Ensure these options collapse when the user navigates past them.


sathomas:

PR #3941, opened 2015-06-19
(A11y severity 3) Remove redundant title attribute from grid items

sathomas:
The headings in the file boxes, "Test discussion" in this example, have `title` text that is the same as the link text. Some screen readers read both `title` text as well as the visible text causing redundant information to be presented to the user. Since `title` text is generally meant for advisory information and no new information is presented, it should be removed wherever it matches the visible text.

### DIFF
--- a/node_modules/oae-admin/tenants/tenants.html
+++ b/node_modules/oae-admin/tenants/tenants.html
@@ -54,7 +54,7 @@
                                <div class="jeditable-field" data-alias="${tenant.alias|encodeForHTMLAttribute}" data-field="host">${tenant.host|encodeForHTML}</div>
                             </td>
                         {else}
-                            <td><a href="//${tenant.host}" title="${tenant.displayName|encodeForHTMLAttribute}" target="_blank">${tenant.host|encodeForHTML}</a></td>
+                            <td><a href="//${tenant.host}" target="_blank">${tenant.host|encodeForHTML}</a></td>
                         {/if}
                         <td>
                             {if tenant.active}
--- a/node_modules/oae-admin/tenants/tenants.html
+++ b/node_modules/oae-admin/tenants/tenants.html
@@ -54,7 +54,11 @@
                                <div class="jeditable-field" data-alias="${tenant.alias|encodeForHTMLAttribute}" data-field="host">${tenant.host|encodeForHTML}</div>
                             </td>
                         {else}
-                            <td><a href="//${tenant.host}" target="_blank">${tenant.host|encodeForHTML}</a></td>
+                            {if tenant.displayName === tenant.host}
+                                <td><a href="//${tenant.host}" target="_blank">${tenant.host|encodeForHTML}</a></td>
+                            {else}
+                                <td><a href="//${tenant.host}" title="${tenant.displayName|encodeForHTMLAttribute}" target="_blank">${tenant.host|encodeForHTML}</a></td>
+                            {/if}
                         {/if}
                         <td>
                             {if tenant.active}
--- a/node_modules/oae-admin/tenants/tenants.html
+++ b/node_modules/oae-admin/tenants/tenants.html
@@ -54,11 +54,7 @@
                                <div class="jeditable-field" data-alias="${tenant.alias|encodeForHTMLAttribute}" data-field="host">${tenant.host|encodeForHTML}</div>
                             </td>
                         {else}
-                            {if tenant.displayName === tenant.host}
-                                <td><a href="//${tenant.host}" target="_blank">${tenant.host|encodeForHTML}</a></td>
-                            {else}
-                                <td><a href="//${tenant.host}" title="${tenant.displayName|encodeForHTMLAttribute}" target="_blank">${tenant.host|encodeForHTML}</a></td>
-                            {/if}
+                            <td><a href="//${tenant.host}" title="${tenant.displayName|encodeForHTMLAttribute}" target="_blank">${tenant.host|encodeForHTML}</a></td>
                         {/if}
                         <td>
                             {if tenant.active}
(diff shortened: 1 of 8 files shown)


sathomas:
There weren't too many cases in which the title was strictly redundant. There were cases in which it was semi-redundant. Mostly those cases were buttons with a title and text content that duplicated the title; however the button content also included an icon. I left the semi-redundant cases as-is, since that might help sighted users discover the icon meaning (e.g. by hovering). To @nicolaasmatthijs for review.

shared/oae/macros/list.html
```diff
@@ -332,7 +332,7 @@
     {/if}
 
     {if profilePath && displayOptions.addLink !== false}
-        <h3 {if displayOptions.threedotTitle !== false} class="oae-threedots"{/if}><a href="${profilePath|profilePath}" title="${entityData.displayName|encodeForHTMLAttribute}" {if displayOptions.linkTarget} target="${displayOptions.linkTarget}"{/if}>${entityData.displayName|encodeForHTML}</a></h3>
```
nicolaasmatthijs: This title attribute isn't entirely redundant. As threedotting is applied here, the tooltip is the only way in which the full title can be seen.

node_modules/oae-core/aboutcontent/aboutcontent.html
```diff
@@ -53,7 +53,7 @@ <h3 id="aboutcontent-title">${contentProfile.displayName|encodeForHTML}</h3>
             </div>
             {if contentProfile.resourceSubType === 'link'}
                 <div id="aboutcontent-link">
-                    <a href="${contentProfile.link}" title="${contentProfile.displayName|encodeForHTMLAttribute}" target="_blank">${contentProfile.link|encodeForHTML}</a>
+                    <a href="${contentProfile.link}" target="_blank">${contentProfile.link|encodeForHTML}</a>
```
nicolaasmatthijs: This title attribute isn't entirely redundant as it provides a readable tooltip for the link

node_modules/oae-admin/tenants/tenants.html
```diff
@@ -54,7 +54,7 @@
                                <div class="jeditable-field" data-alias="${tenant.alias|encodeForHTMLAttribute}" data-field="host">${tenant.host|encodeForHTML}</div>
                             </td>
                         {else}
-                            <td><a href="//${tenant.host}" title="${tenant.displayName|encodeForHTMLAttribute}" target="_blank">${tenant.host|encodeForHTML}</a></td>
```
nicolaasmatthijs: This title attribute isn't entirely redundant as it provides a readable tooltip for the link

nicolaasmatthijs:

sathomas:
As best as I can figure, we have a couple of options here:

1. Close this issue as `WONTFIX`.
2. Instead of using the `title` attribute, use, e.g., the bootstrap [tooltip](http://getbootstrap.com/javascript/#tooltips) component.

The second choice offers the best screen reader experience since it eliminates the duplicate spoken text. Is that your preference @nicolaasmatthijs or would you prefer the simpler implementation that sacrifices some accessibility?

nicolaasmatthijs:
I don't think this issue needs to be closed as `WONTFIX`. There are a number of cases in the PR where we have a genuinely redundant title attribute, and those can be removed. For those where the title attribute is not entirely redundant, I would prefer to keep the existing title attributes in place.

sathomas:
followed up on review; back to @nicolaasmatthijs

node_modules/oae-admin/tenants/tenants.html
```diff
@@ -54,7 +54,11 @@
...
+                            {if tenant.displayName === tenant.host}
```
nicolaasmatthijs: The tenant displayname is never going to be the same as the tenant host. We can just always have the title attribute here

node_modules/oae-core/aboutcontent/aboutcontent.html
```diff
@@ -53,7 +53,11 @@ <h3 id="aboutcontent-title">${contentProfile.displayName|encodeForHTML}</h3>
             </div>
             {if contentProfile.resourceSubType === 'link'}
                 <div id="aboutcontent-link">
-                    <a href="${contentProfile.link}" title="${contentProfile.displayName|encodeForHTMLAttribute}" target="_blank">${contentProfile.link|encodeForHTML}</a>
```
nicolaasmatthijs: It might be better to do:

```
<a href="${contentProfile.link}" {if contentProfile.displayName !== contentProfile.link} title="${contentProfile.displayName|encodeForHTMLAttribute}" {/if} target="_blank">${contentProfile.link|encodeForHTML}</a>
```

So we don't have to repeat the markup of the link

shared/oae/js/jquery-plugins/jquery.infinitescroll.js
```diff
@@ -274,7 +274,20 @@ define(['jquery', 'underscore', 'oae.api.util', 'oae.api.i18n', 'oae.api.l10n'],
                 oaeL10n.timeAgo($listContainer);
 
                 // Apply multi-line threedotting to the tile titles
-                $('.oae-tile h3').dotdotdot({'watch': 'window'});
+                $('.oae-tile h3').dotdotdot({
+                    'callback': function(isTruncated, orgContent) {
```
nicolaasmatthijs: I'd prefer to not have inconsistent tooltip behaviour within the same list. I'd rather sacrifice some accessibility here and always have the title attribute for the list items.

nicolaasmatthijs:

sathomas:
back to @nicolaasmatthijs

shared/oae/macros/list.html
```diff
@@ -332,7 +332,7 @@
...
+        <h3 {if displayOptions.threedotTitle !== false} class="oae-threedots"{/if}><a href="${profilePath|profilePath}" id="${_.uniqueId('oae-')}" title="${entityData.displayName|encodeForHTMLAttribute}" {if displayOptions.linkTarget} target="${displayOptions.linkTarget}"{/if}>${entityData.displayName|encodeForHTML}</a></h3>
```
nicolaasmatthijs: The id can be removed again. I'll follow up

nicolaasmatthijs:
Followed up and merged

PR #3931, opened 2015-06-19
(A11y severity 3) Reorganize headings for accessibility

sathomas:
The majority of the pages in this site have multiple first-level headings (`<h1>`). While the presence of headings is good, pages should generally have a single `h1` for the document title/main heading. This facilitates quick navigation directly to the main content of the page.

### DIFF
--- a/ui/content.html
+++ b/ui/content.html
@@ -38,7 +38,7 @@
                                 ${renderThumbnail(content, displayOptions)}
                                 <div>
                                     <div>
-                                        <h1 class="oae-threedots">${content.displayName|encodeForHTML}</h1>
+                                        <span class="oae-clip-text oae-threedots">${content.displayName|encodeForHTML}</span>
                                         <i class="fa fa-caret-down"></i>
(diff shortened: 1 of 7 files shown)


sathomas:

ui/content.html
```diff
@@ -38,7 +38,7 @@
                                 ${renderThumbnail(content, displayOptions)}
                                 <div>
                                     <div>
-                                        <h1 class="oae-threedots">${content.displayName|encodeForHTML}</h1>
+                                        <span class="oae-clip-text oae-threedots">${content.displayName|encodeForHTML}</span>
```
nicolaasmatthijs: I consider the primary clip title to be the main title for the page (user name, group name, content item name, etc.) and think that it should remain an `h1`.

nicolaasmatthijs: I agree that the titles in the secondary clips should not be `h1`s though.

nicolaasmatthijs:

sathomas:
back to @nicolaasmatthijs

PR #3934, opened 2015-06-23
(A11y severity 3) Align navigation and visual order

sathomas:
Multiple places throughout the site have a navigation order that does not follow the visual order (generally wherever usernames are concerned - e.g., "Share With", "Notifications", and "Comments" sections). Ensure that page navigation follows a logical left to right order.


sathomas:
over to @nicolaasmatthijs for review

PR #4031, opened 2015-06-24
Email verification

mrvisser:
* When the user updates their email address in editprofile, an email verification is sent to the new email address
* Rather than closing the editprofile modal when the user has updated their email address, we navigate the modal to a panel that instructs them that an email has been sent and they need to verify it before they can use it
* The "force complete profile" modal has now been switched over to the editprofile modal so it has the same email verification functionality. The user cannot close the editprofile modal until they have a valid profile
* When there is a `/?url=something` query string parameter, the topnavigation will open up an appropriate login action for you. Synonymous to clicking on the "Sign In" at the top right

Depends on:
 - https://github.com/oaeproject/Hilary/pull/1203


mrvisser:

mrvisser:
Note that this doesn't work well w/ ie9 and ie10 yet, but I'd like a bit of design review before I start getting into that nitty gritty.

nicolaasmatthijs:
Closing in favour of https://github.com/oaeproject/3akai-ux/pull/4051

PR #4033, opened 2015-06-26
["Manage access" popup] "Add" button is not clickable on Internet Explorer 11

sathomas:
When using Internet Explorer 11, the "Add" button is disabled on the *popup* when trying to share a document with other users/groups.
oae-error_ui

The issue may be related to the use of the remove() method which is apparently not supported in IE:  
error-oae


sathomas:

nicolaasmatthijs:
Looks good. Merged.

PR #4034, opened 2015-07-06
Sign Up modal doesn't allow email address as username

sathomas:
The Sign Up modal doesn't allow an email address as the username because the `@` is considered to be an invalid character. However, we want to allow email addresses as usernames and the REST API already supports this.


sathomas:

nicolaasmatthijs:
Looks good. Merged.

PR #4036, opened 2015-07-13
Allow email notifications for user to be turned off in admin UI

sathomas:
The OAE back-end has an option to turn off email notifications for a user, but it’s currently not exposed in the UI to discourage users from turning them off. You can currently turn off email notifications through the REST API and the Swagger documentation page, but we should also expose this as a fourth email notification option in the manageuser widget.

The fa-ban icon can be used for the new option.

### DIFF
--- a/node_modules/oae-admin/manageuser/manageuser.html
+++ b/node_modules/oae-admin/manageuser/manageuser.html
@@ -162,7 +162,7 @@
     <div id="manageuser-email-options-container">
         <div id="manageuser-email-options" class="clearfix">
             <div class="oae-large-options-container row">
-                <div class="col-sm-4 text-center">
+                <div class="col-sm-3 text-center">
                     <label for="manageuser-email-immediate" {if userProfile.emailPreference === 'immediate'} class="checked"{/if}>
                         <i class="fa fa-check hide"></i>
                         <div class="well oae-large-options">
@@ -172,7 +172,7 @@
                         </div>
                     </label>
                 </div>
-                <div class="col-sm-4 text-center">
+                <div class="col-sm-3 text-center">
                     <label for="manageuser-email-daily" {if userProfile.emailPreference === 'daily'} class="checked"{/if}>
                         <i class="fa fa-check hide"></i>
                         <div class="oae-large-options well">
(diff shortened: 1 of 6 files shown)


sathomas:

node_modules/oae-admin/manageuser/manageuser.html
```diff
@@ -192,6 +192,16 @@ <h3 id="manageuser-editprofile-header-container"><!-- --></h3>
                         </div>
                     </label>
                 </div>
+                <div class="col-sm-3 text-center">
+                    <label for="manageuser-email-never" {if userProfile.emailPreference === 'never'} class="checked"{/if}>
+                        <i class="fa fa-check hide"></i>
+                        <div class="oae-large-options well">
+                            <input type="radio" id="manageuser-email-never" value="never" name="manageuser-email-group" class="pull-left" tabindex="0" {if userProfile.emailPreference === 'never'} checked{/if}/>
+                            <i class="fa fa-ban large text-center {if userProfile.emailPreference === 'weekly'} selected{/if}"></i>
```
nicolaasmatthijs: This needs to be `never`

nicolaasmatthijs: Interestingly enough, I don't see a visual difference between applying `selected` to the icon and not applying it. Can you try to find if it still does anything anywhere and if not, can you go ahead and remove the `if` statement in all places where we use the `oae-large-options` ?

nicolaasmatthijs:
Left a comment. Back to @sathomas for follow-up

node_modules/oae-admin/manageuser/manageuser.html
sathomas: No difference. The only rule that applies to `selected` (setting the color) is also defined to apply when the corresponding `<input>` is checked. In `skin.css`:

```
.fa.active, /* ... */  .fa.selected, /* ... */ input[type=radio]:checked+.fa {
    color: #555
}
```

Looks like the only other place it's used is in the preferences widget. I'll remove it there as well.

sathomas:
back to @nicolaasmatthijs after followup

node_modules/oae-core/preferences/preferences.html
```diff
@@ -111,7 +111,7 @@ <h3 id="preferences-modal-title">__MSG__MY_PREFERENCES__</h3>
                             <i class="fa fa-check hide"></i>
                             <div class="well oae-large-options">
                                 <input type="radio" id="preferences-email-immediate" value="immediate" name="preferences-email-group" class="pull-left" tabindex="0" {if oae.data.me.emailPreference === 'immediate'} checked{/if}/>
-                                <i class="fa fa-bolt large text-center {if oae.data.me.emailPreference === 'immediate'} selected{/if}"></i>
```
nicolaasmatthijs: It looks like the `selected` class is also being applied through js in the following cases:

```
node_modules/oae-admin/manageuser/js/manageuser.js:                $('.oae-large-options i', $optionsContainer).removeClass('selected');
node_modules/oae-admin/manageuser/js/manageuser.js:                $(this).closest('label').find('.oae-large-options i').addClass('selected');
node_modules/oae-core/preferences/js/preferences.js:                $('.oae-large-options i', $rootel).removeClass('selected');
node_modules/oae-core/preferences/js/preferences.js:                $(this).parents('label').find('.oae-large-options i').addClass('selected');
```

nicolaasmatthijs: Can you verify that removing the selected styling is not causing any visual regressions in those areas either?

nicolaasmatthijs:
Left another comment. Back to @sathomas for follow-up

sathomas:
Looks safe to remove the JavaScript code. Back to @nicolaasmatthijs

nicolaasmatthijs:
Looks good. Merged

PR #3992, opened 2015-07-15
Top navigation refactoring

sathomas:
- Remove small logo
- Provide equal icon spacing on small devices
- Top navigation search bar should be icon only when the user is on the search page
- Refactor signed out top navigation


sathomas:
It's hard to see from the low resolution design comps, but it looks like there are other design changes in addition to those listed on the issue bullet list (e.g. removing margins entirely on small screen). I didn't make any of those changes, but I think I've got all the bullet points. To @nicolaasmatthijs for review.

nicolaasmatthijs:
I'm coordinating with Sam around a few questions on how certain things should be behaving. In the meantime, I do have a couple of comments:

- There's a merge conflict
- When the top navigation is in authenticated mode on a small device, the icons should be evenly spaced. Currently, the space between the second and third icon is larger than the space between the other icons.

screen shot 2015-07-22 at 22 55 26

- The evenly spaced icons should no longer be used at the 768 breakpoint. At that point, the logo should be shown and the icons should move to the left and right again

nicolaasmatthijs:

sathomas:
I __think__ this has the latest comments addressed, but I'm not certain I fully understand the desired behavior. In any case, I'll assign this to @nicolaasmatthijs until he and @SamPeck have a chance to consult.

node_modules/oae-core/register/register.html
```diff
@@ -9,7 +9,7 @@
                 <button type="button" class="close" data-dismiss="modal" aria-label="__MSG__CLOSE__">
```
nicolaasmatthijs: After lots of back and forth around the issue of the logo on the left hand side clashing with the primary clip, we've decided to go back to the following scenario:

- The anonymous header should stay as is with `Sign up` on the left (when applicable) and `Search` and `Sign in` on the right.
- The search box should be a search icon on small devices
- Search should be hidden on the landing page and search page
- The anonymous header should not have a small logo

For authenticated users, we do want to retain the changes made in this PR. The only difference is that on wider screens, the search box should be a search icon on the landing page and search page.

node_modules/oae-core/topnavigation/css/topnavigation.css
```diff
@@ -32,16 +32,37 @@
...
+    }
+    #topnavigation-container.topnavigation-search #topnavigation-search-form {
         display: none;
     }
 }
 
+/* iPad resolution */
+@media (max-width : 767px) {
+    #topnavigation-left.topnavigation-icons ul li:last-of-type {
+        left: calc(67% - 70px);
```
nicolaasmatthijs: Can we come up with an approach that doesn't require `calc`?

node_modules/oae-core/topnavigation/topnavigation.html
```diff
@@ -41,10 +21,26 @@ <h1 id="topnavigation-target" class="sr-only">__MSG__SKIP_MENU_TARGET__</h1>
 //////////////////////////////
 -->
 
+<div id="topnavigation-anonymous-template"><!--
```
nicolaasmatthijs: Would it make sense to move the anonymous template into a single template rather than having subtemplates? The same is true for the logged in template.

shared/oae/css/oae.skin.less
```diff
@@ -71,18 +71,11 @@ body {
 
 /* Institutional logo */
 @institutional-logo-url: '/shared/oae/img/oae-logo.png';
-/* Small institutional logo */
-@institutional-logo-small-url: '/shared/oae/img/oae-logo-small.png';
```
nicolaasmatthijs: Can the small logo file be removed as well?

nicolaasmatthijs:
Finished another round of review. Back to @sathomas for follow-up

nicolaasmatthijs:
This is now part of https://github.com/oaeproject/3akai-ux/pull/4040

PR #4037, opened 2015-07-20
Server maintenance page is blank

sathomas:
Transferred from https://github.com/oaeproject/puppet-hilary/issues/163#event-360029992

### DIFF
--- a/shared/oae/errors/noscript.html
+++ b/shared/oae/errors/noscript.html
@@ -21,7 +21,7 @@
         </script>
 
     </head>
-    <body style="display: block;">
+    <body style="display: block; visibility: visible;">
         <!-- BODY -->
         <main class="oae-branding-container error-static oae-main-content">
             <div id="error-content">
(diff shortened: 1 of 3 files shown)


sathomas:

sathomas:
BTW, an alternative approach would be to add

    body {
        visibility: visible;
    }

in the `oae.error.css` file since (at least currently) that file is always included __after__ `oae.core.css`. But doing so would make CSS inclusion order-dependent, which seems risky. (It would also change the scope of `oae.error.css` as that file currently only styles error elements.)

We could also add `!important` to avoid ordering problems if we wanted to rely on `!important`.

shared/oae/errors/noscript.html
```diff
@@ -21,7 +21,7 @@
         </script>
 
     </head>
-    <body style="display: block;">
+    <body style="display: block; visibility: visible;">
```
nicolaasmatthijs: It looks like this regression was introduced as part of https://github.com/oaeproject/3akai-ux/commit/713a6898dffd85fe7a8c7aeeefeb5c90949cc114. Adding `visibility: visible` to the `body` tags should be fine. As we have switched from display to visibility, the `display: block` styles can be removed now. I'll push this as follow-up.

nicolaasmatthijs:
Followed up and merged

PR #3943, opened 2015-07-22
(A11y severity 3) Improve structure of activity stream content

sathomas:
The content within the "Recent Activity" section could benefit from a better heading structure. For example, the first line at the start of each activity should be marked up as a second-level heading (it is currently not a heading). This will allow users to more effectively navigate this page.

### DIFF
--- a/node_modules/oae-core/activity/css/activity.css
+++ b/node_modules/oae-core/activity/css/activity.css
@@ -20,9 +20,9 @@
 }
 
 .activity-widget .activity-summary {
-	font-size: 14px;
-	font-weight: normal;
-	margin: 0;
+    font-size: 14px;
+    font-weight: normal;
+    margin: 0;
     padding-left: 50px;
     word-wrap: break-word;
 }
@@ -33,7 +33,7 @@
 
 .activity-widget .activity-summary > small {
     display: block;
-	line-height: 1.5;
+    line-height: 1.5;
     margin-top: 2px;
 }
 
--- a/node_modules/oae-core/activity/css/activity.css
+++ b/node_modules/oae-core/activity/css/activity.css
@@ -20,6 +20,9 @@
 }
 
 .activity-widget .activity-summary {
+	font-size: 14px;
+	font-weight: normal;
+	margin: 0;
     padding-left: 50px;
     word-wrap: break-word;
 }
@@ -30,6 +33,7 @@
 
 .activity-widget .activity-summary > small {
     display: block;
+	line-height: 1.5;
     margin-top: 2px;
 }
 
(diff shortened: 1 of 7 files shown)


sathomas:
I think this approach might be a reasonable compromise. To @nicolaasmatthijs for review

nicolaasmatthijs:
@sathomas : This approach seems reasonable to be.

However, I was wondering if it semantically makes sense for the list item tiles to be an `h3`. They don't strictly correspond with headers in the page, and having the `h3` be part of the list macro makes ensuring that we have the header order right a bit tricky. If having them as a header makes sense, then I'll go ahead and merge this PR. If it doesn't really make sense, perhaps we can change the `h3` to a `span`?

nicolaasmatthijs:
Back to @sathomas for feedback

sathomas:
@nicolaasmatthijs I think you're right in that using `h3` within the list macro is problematic. Since the macro can theoretically be used anywhere, it can't really rely on any heading hierarchy. I wasn't sure if we wanted to make a change to the list macro since the effects could be widespread, but I'll take a look. Perhaps it won't be too disrupting to change the heading to a span. We can still use aria attributes within the macro to make it's internal hierarchy clear to assistive technology.

At the risk of opening a larger can of worms, @mrvisser, @simong, @stuartf, are there any other tweaks to the list macro that folks have wanted to see? If so, I can take a look at those at the same time.

mrvisser:
@sathomas none come to mind

sathomas:
I'm sure I've missed some required changes, but tiles now use `oae-tile-title` class instead of `<h3>` so the `/me` heading hierarchy is saner. To @nicolaasmatthijs for initial review

node_modules/oae-core/activity/css/activity.css
```diff
@@ -20,6 +20,9 @@
 }
 
 .activity-widget .activity-summary {
+	font-size: 14px;
```
nicolaasmatthijs: Indentation is off

shared/oae/css/oae.components.css
```diff
@@ -972,9 +972,9 @@ div.oae-thumbnail i.fa {
  *      </a>
  *      <div class="oae-tile-metadata">
  *          <i class="fa fa-oae-public pull-left"></i>
- *          <h3 class="oae-threedots">
+ *          <div class="oae-tile-title oae-threedots">
```
nicolaasmatthijs: Rather than requiring another class in this "component", maybe we can just wrap the title in a `span` and create the CSS rules against the `span` directly

nicolaasmatthijs: Your call if you want to make this change or not

nicolaasmatthijs:
Left some minor review. Back to @sathomas for follow-up

shared/oae/css/oae.components.css
sathomas: Targeting naked elements with style rules tends to make me nervous. Too easy (at least for me) to mess things up later.

sathomas:
I left the class in for now. To @nicolaasmatthijs

nicolaasmatthijs:
Looks good. Merged

PR #4054, opened 2015-09-09
Auto-focus in search box on search page

mrvisser:
When the search page is loaded, focus should automatically be set on the search box. Note that the cursor should be set at the end of the text in the field.


mrvisser:
Looks like the `oae.api.js` initialization process can wreck autofocus, so I just restore it after initial widgets are loaded

mrvisser:

shared/oae/api/oae.api.js
```diff
@@ -132,6 +132,11 @@ define(['oae.api.admin', 'oae.api.authentication', 'oae.api.config', 'oae.api.co
                                         // initial widget loading have finished
                                         $('body').css('visibility', 'visible');
 
+                                        // Apply auto-focus after the core HTML widgets have loaded.
```
nicolaasmatthijs: Neat solution

shared/oae/api/oae.api.js
```diff
@@ -132,6 +132,11 @@ define(['oae.api.admin', 'oae.api.authentication', 'oae.api.config', 'oae.api.co
...
+                                        // The rendering will cause HTML5 autofocus to be lost, so
+                                        // we re-inforce it here
+                                        $('[autofocus]:first').focus();
```
nicolaasmatthijs: I always get the cursor at the beginning of the text field. The cursor should instead be positioned at the end of the value of the text field.

nicolaasmatthijs:

mrvisser:
It was intermittent. If the value gets set in search.js before the autofocus happens, then the focus would put the cursor to the beginning.

mrvisser:
Back to @nicolaasmatthijs

nicolaasmatthijs:
Looks good. Merged.

PR #4061, opened 2015-09-19
(.gitignore) Add workaround for Atom tree-view .gitignore support issue

mrvisser:


simong:
Excuse the noise, trying to help sam get his notifications sorted so he can jump in when highlighted.

@SamPeck or sam never replies to any mentions for issues in this repository because emails that include his name are buried in all the other Github notifications and other emails. While also not being an active member of the oaeproject repository won't let third party notification services allow him to connect to it. This is a test to see if the Slack integration can take care of this problem.

nicolaasmatthijs:
Looks good. Merged.

PR #4066, opened 2015-09-22
(InvitingGuests) Improper alignment in mobile devices

mrvisser:
I used android device and firefox browser to test the page, I found that the icons like home,notifications, search and logout  are not aligned properly. when i clicked on "notifications', it got opened but it got pointed from 'home' icon.
screenshot_2015-09-22-12-17-55


mrvisser:

node_modules/oae-core/notifications/css/notifications.css
```diff
@@ -102,7 +107,7 @@
     }
 }
 
-@media (max-width: 480px) {
+@media (max-width: 479px) {
```
nicolaasmatthijs: Positioning is still a bit off on the smallest size:

screen shot 2015-09-22 at 21 43 17

nicolaasmatthijs:

node_modules/oae-core/notifications/css/notifications.css
mrvisser: I missed that the icons will eventually fluidly squish together. We can't do this since the popover needs fixed positioning.

I've made the icons a bit less "centered" along the top in unconventional widths.

When the badge was showing, it was also giving more width to the megaphone, and since it is centered it was causing the position of the megaphone to change (in production, the button drives the width of the `li`, so it wasn't an issue). To fix this, I take the badge off the baseline so it no longer impacts the width of the button -- I think this is a bit more stable approach than what we relied before, because the widths of those buttons will always be their non-badge version that is most commonly tested with and should never result in wrapping if a test case is missed.

mrvisser:
Back to @nicolaasmatthijs

nicolaasmatthijs:
Looks good. Merged.

PR #4073, opened 2015-09-25
Two modals display if accepting invitation, don't have email (signin with twitter) and need to accept T&C

mrvisser:
image

I think the POST to accept invitations triggers a T&C modal while the email modal is showing.

Additionally, I'm not sure if the accept invitation request is going to work anyway with external login when T&C is needed as T&C needs to be accepted before the invitation.

### DIFF
--- a/node_modules/oae-core/termsandconditions/js/termsandconditions.js
+++ b/node_modules/oae-core/termsandconditions/js/termsandconditions.js
@@ -21,6 +21,27 @@ define(['jquery', 'oae.core'], function($, oae) {
         var $rootel = $('#' + uid);
 
         /**
+         * Accept the invitation in the URL state, if any
+         */
+        var acceptInvitation = function() {
+            var invitationToken = oae.api.util.url().param('invitationToken');
+            if (!invitationToken) {
+                // Do not continue if there is no invitation info
+                return;
+            }
+
+            // Accept the invitation
+            oae.api.user.acceptInvitation(invitationToken, function(err, result) {
+                if (err && err.code !== 404) {
+                    oae.api.util.notification(
+                        oae.api.i18n.translate('__MSG__EMAIL_INVITATION_FAILED__'),
+                        oae.api.i18n.translate('__MSG__AN_ERROR_OCCURRED_WHILE_ACCEPTING_YOUR_INVITATION__'),
+                        'error');
+                }
+            });
+        };
+
+        /**
          * Fetch the Terms and Conditions and render the parsed markdown into the widget.
          */
         var renderTermsAndConditions = function() {
@@ -47,6 +68,10 @@ define(['jquery', 'oae.core'], function($, oae) {
                     $('#termsandconditions-modal', $rootel).modal('unlock');
                     // Hide the modal when the Terms and Conditions have been accepted
                     $('#termsandconditions-modal', $rootel).modal('hide');
+
+                    // Accept any resource invitations that we have delayed to
+                    // show the T&C
+                    acceptInvitation();
                 } else {
                     oae.api.util.notification(
                         oae.api.i18n.translate('__MSG__TERMS_AND_CONDITIONS_NOT_ACCEPTED__', 'termsandconditions'),
--- a/shared/oae/api/oae.api.user.js
+++ b/shared/oae/api/oae.api.user.js
@@ -259,6 +259,10 @@ define(['exports', 'jquery', 'underscore', 'oae.api.config'], function(exports, 
                 'token': token
             },
             'success': function(data) {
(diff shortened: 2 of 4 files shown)


mrvisser:
Test cases:

1. User has invalid profile, needs to accept T&C and is following an invitation
  * Setup: Need to set the `displayName` to `null` in Cassandra using cqlsh and clear redis, then accept an invitation as that user
  * Expected workflow:
    1. Login as the user following an invitation
    2. User gets the edit profile modal, they must put in their display name and Save
    3. Edit profile modal closes, T&C opens. User must accept & continue
    4. T&C modal closes, invitations are then automatically accepted and the activity feed begins to fill up
2. User has invalid profile, is following an invitation
  * Setup: Need to set the `displayName` to `null` in Cassandra using cqlsh and clear redis, then accept an invitation as that user
  * Expected workflow:
    1. Login as the user following an invitation
    2. User gets the edit profile modal and invitation is accepted concurrently
    3. When the user fills in their profile details and saves, the activities are waiting for them in their feed
3. User needs to accept T&C, has valid profile, is following an invitation
  * Setup: Enable T&C
  * Expected workflow:
    1. Login as the user (e.g., using twitter, don't create an account and accept T&C) following an invitation
    2. User gets the T&C, accepts the T&C
    3. Modal closes and then the invitations are accepted and the activity feed populates on the fly

Note that when https://github.com/oaeproject/Hilary/pull/1238 is merged, there should be no case when a user follows an invitation and does not have a verified email address when they create their account.

shared/oae/api/oae.api.user.js
```diff
@@ -259,6 +259,10 @@ define(['exports', 'jquery', 'underscore', 'oae.api.config'], function(exports,
                 'token': token
             },
             'success': function(data) {
+                // Accepting an invitation will set a verified email if there wasn't one already,
+                // so do that here to avoid giving the user a pop-up
+                require('oae.core').data.me.email = require('oae.core').data.me.email || data.email;
```
nicolaasmatthijs: In what scenario do we need to use the existing email from the me object instead of the one returned in the data object?

node_modules/oae-core/termsandconditions/js/termsandconditions.js
```diff
@@ -21,6 +21,27 @@ define(['jquery', 'oae.core'], function($, oae) {
         var $rootel = $('#' + uid);
 
         /**
+         * Accept the invitation in the URL state, if any
+         */
+        var acceptInvitation = function() {
+            var invitationToken = oae.api.util.url().param('invitationToken');
```
nicolaasmatthijs: Does this mean that invitation won't be accepted in the following scenario:

1. The user accepts their invitation in a way that doesn't supply a confirmed email address
2. The user enters their email address
3. The user clicks the verify link in the verify email
4. The user is now shown the T&C modal, but the `invitationToken` is no longer present in the URL because they followed the verify link

nicolaasmatthijs:

node_modules/oae-core/termsandconditions/js/termsandconditions.js
mrvisser: As of the Hilary ticket that was merged, there is no way to accept an invitation and not result in a verified email address.

nicolaasmatthijs: So what if you follow the invitation and then

a. Sign in with Twitter (which doesn't release an email address)
b. Sign in with Shibboleth/CAS that doesn't release a (valid) email address

nicolaasmatthijs: Oh, it then uses the email address you were invited with?

shared/oae/api/oae.api.user.js
mrvisser: In scenarios where we expect the `oae.data.me` to represent the latest version of the user profile outside of this process. Such as in `editprofile.js` to fill out the initial profile information of the form, `oae.api.js` where we determine if the user has a verified email, etc...

node_modules/oae-core/termsandconditions/js/termsandconditions.js
mrvisser: That's correct. This is the change the Hilary ticket makes.

https://github.com/oaeproject/Hilary/pull/1238/files#diff-ec09272681a929e971f8efad444b685eR246 sets the invitation token in the create-user process if there is no email provided by external auth.

And, the authentication API already sets the email associated with the invitation token if there is no email set in the create user request (i.e., if you login VIA local auth).

nicolaasmatthijs:
All questions have been addressed by @mrvisser . Merged.

PR #4075, opened 2015-09-29
Cannot confirm email unless you are already signed in

mrvisser:
When the user is not already signed in to the OAE and they click on the "Verify my email" button in the email verification email, they are not presented the signin option when they arrive at the OAE, and there is no longer any redirect url to accept the verification.

Previously, redirecting to `/me` would cause a redirect to `/` with a `?url=` parameter for redirect. We should try and do something similar on the `/` page, however it will be a bit trickier because we need to reliably determine when to redirect. If we just send directly to `/?url=...`, then that might not have the desired effect when the user is already signed in.

### DIFF
--- a/shared/oae/api/oae.api.js
+++ b/shared/oae/api/oae.api.js
@@ -180,18 +180,16 @@ define(['underscore', 'oae.api.admin', 'oae.api.authentication', 'oae.api.config
                 return;
             }
 
-            // Perform any invitation accepting if instructed and if possible
-            acceptInvitation(function() {
+            // Perform any email verification if instructed before we try and determine if the
+            // user's profile info is valid
+            verifyEmail(function() {
+                // No other pre-use actions are applicable to anonymous users
+                if (oae.data.me.anon) {
+                    return;
+                }
 
-                // Perform any email verification if instructed before we try and determine if the
-                // user's profile info is valid
-                verifyEmail(function() {
-                    // Do not make an anonymous user clean up their profile or
-                    // accept T&C
-                    if (oae.data.me.anon) {
-                        return;
-                    }
-
+                // Perform any invitation accepting if instructed and if possible
+                acceptInvitation(function() {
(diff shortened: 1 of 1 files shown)


mrvisser:

mrvisser:
@nicolaasmatthijs ideally we could get this fix through in time for the bug bash today

shared/oae/api/oae.api.js
```diff
@@ -191,6 +186,12 @@ define(['underscore', 'oae.api.admin', 'oae.api.authentication', 'oae.api.config
                 // Perform any email verification if instructed before we try and determine if the
                 // user's profile info is valid
                 verifyEmail(function() {
```
nicolaasmatthijs: Can we swap `acceptInvitation` and `verifyEmail` in the order? In that case, we wouldn't have to do 

```
// We cannot accept an invitation as the anonymous user. Let the
// signup page handle it
if (oae.data.me.anon) {
  return callback();
}
```

in `acceptInvitation`

shared/oae/api/oae.api.js
```diff
@@ -212,6 +213,12 @@ define(['underscore', 'oae.api.admin', 'oae.api.authentication', 'oae.api.config
          * @param  {Function}       callback            Invoked when the accept invitation request is complete
          */
         var acceptInvitation = function(callback) {
+            // We cannot accept an invitation as the anonymous user. Let the
+            // signup page handle it
+            if (oae.data.me.anon) {
```
nicolaasmatthijs: Don't we also need to redirect to login when there is an invitation token ?

nicolaasmatthijs:

shared/oae/api/oae.api.js
mrvisser: Incoming invitations are sent to the `/signup` page, which handles this specially. There should be no case to be on `/`, for example, and be anonymous with invitation data in the query string.

mrvisser:

nicolaasmatthijs:
Looks good. Merged

PR #4077, opened 2015-10-03
Ensure a user enters an email that belongs to the tenant’s configured email domain

simong:


shared/oae/bundles/ui/default.properties
```diff
@@ -587,6 +587,7 @@ TEXT_DOCUMENT = Text document
 THE_PASSWORD_SHOULD_BE_AT_LEAST_SIX_CHARACTERS_LONG = The password should be at least 6 characters long
 THE_USERNAME_SHOULDNT_CONTAIN_SPACES = The username shouldn't contain spaces
 THE_USERNAME_SHOULD_BE_AT_LEAST_THREE_CHARACTERS_LONG = The username should be at least 3 characters long
+THIS_EMAIL_DOMAIN_IS_NOT_ALLOWED = This email domain is not allowed
```
nicolaasmatthijs: Something along the lines of `You can only register with an 'ox.ac.uk' email address` would be friendlier

ui/js/signup.js
```diff
@@ -199,6 +202,13 @@ require(['jquery', 'underscore', 'oae.core'], function($, _, oae) {
             return;
         }
 
+        // Get the email domain for the tenant (if any) so we can validate the email
+        oae.api.admin.getTenant(null, function(err, tenant) {
```
nicolaasmatthijs: Is there a way in which we can avoid doing another API call? Can we return the email domain as part of the tenant object in the `me` feed?

ui/js/signup.js
```diff
@@ -199,6 +202,13 @@ require(['jquery', 'underscore', 'oae.core'], function($, _, oae) {
...
+            if (!err && tenant.emailDomain) {
+                emailDomain = new RegExp(tenant.emailDomain + '$');
```
nicolaasmatthijs: Wouldn't this cause `bla@fooox.ac.uk` to be a valid address when `ox.ac.uk` is configured as the email domain?

ui/js/signup.js
```diff
@@ -239,6 +249,9 @@ require(['jquery', 'underscore', 'oae.core'], function($, _, oae) {
                 },
                 'password_repeat': {
                     'equalTo': '#signup-createaccount-password'
+                },
+                'email': {
+                    'emaildomain': true
```
nicolaasmatthijs: Rather than adding this here, you can just add `emaildomain` as a property to the email field

nicolaasmatthijs:
Finished review. Back to @simong for follow-up

ui/js/signup.js
mrvisser: Also, the `.`'s would be treated as "any character" instead of actual dots since they are unescaped in the regexp.

In the past on the server side, I've checked to see if it is exactly equal, if not, check that it ends with the tenant email domain preceded with a period. If not, then it doesn't match:

e.g., `'caret.cam.ac.uk' === 'cam.ac.uk'`? No. `'caret.cam.ac.uk'.endsWith('.cam.ac.uk')` ? Yes. So it's valid.

simong:
Pushed some follow-up

ui/signup.html
```diff
@@ -60,7 +60,7 @@ <h3 id="signup-title-container"><!----></h3>
                         <h4>__MSG__CREATE_A_NEW_ACCOUNT_COLON__</h4>
                         <form id="signup-createaccount-form" class="form-horizontal" role="form">
                             {if invitationInfo.email}
-                                <input type="hidden" id="signup-createaccount-email" name="email" class="form-control required email maxlength-short" value="${invitationInfo.email|encodeForHTMLAttribute}" />
+                                <input type="hidden" id="signup-createaccount-email" name="email" class="form-control required email maxlength-short emaildomain" value="${invitationInfo.email|encodeForHTMLAttribute}" />
```
nicolaasmatthijs: Do we want all of the validation rules on the invitation email? If the validation fails for some reasons (e.g. domain was changed), the user would have no idea as to why it fails

ui/js/signup.js
```diff
@@ -291,6 +294,29 @@ require(['jquery', 'underscore', 'oae.core'], function($, _, oae) {
...
+
+                        // If the tenant has not been configured with an email domain, any email address
+                        // can be used to sign up
+                        if (!configuredEmailDomain) {
+                            return true;
+                        }
+
+                        // Only accept email addresses that are either an exact match or who have the
+                        // configured email domain as the suffix
+                        var enteredEmailDomain = value.split('@')[1];
```
nicolaasmatthijs: Probably want to use `.pop()` here ?

ui/js/signup.js
```diff
@@ -36,6 +36,9 @@ require(['jquery', 'underscore', 'oae.core'], function($, _, oae) {
     // Variable that specifies if the terms and conditions are enabled
     var termsAndConditionsEnabled = oae.api.config.getValue('oae-principals', 'termsAndConditions', 'enabled');
 
+    // Variable that will hold the regular expression that can be used to validate the email domain
+    var emailDomain = null;
```
nicolaasmatthijs: This is no longer being used

nicolaasmatthijs:
Finished review. Back to @simong for follow-up

simong:
Pushed follow-up

nicolaasmatthijs:
Followed up and merged

shared/oae/bundles/ui/default.properties
```diff
@@ -613,6 +613,7 @@ WHAT_WOULD_YOU_LIKE_TO_CALL_YOUR_GROUP = What would you like to call your group?
 WHAT_WOULD_YOU_LIKE_TO_DISCUSS = What would you like to discuss?
 WORD_DOCUMENT = Word document
 XML_DOCUMENT = XML document
+YOU_CAN_ONLY_REGISTER_WITH_AN_EMAIL_ADDRESS_ENDING_IN_EMAIL_DOMAIN = You can only register with an email address ending in '${emailDomain}'
```
nicolaasmatthijs: I'm going to refine this message a little bit

PR #4078, opened 2015-10-06
Trim AS search query before search

mrvisser:


shared/vendor/js/jquery-plugins/jquery.autoSuggest.js
```diff
@@ -92,6 +93,7 @@
         var request = null;
         if(typeof data == "string" || typeof data == "function") {
             d_fetcher = function(query, next) {
+                query = $.trim(query);
```
nicolaasmatthijs: I have no idea what `data` does underneath and how it feeds into the baseUrl, but the plugin appears to be using query for the query parameter.

nicolaasmatthijs:
Looks good. Merged.

shared/vendor/js/jquery-plugins/jquery.autoSuggest.js
nicolaasmatthijs: Got it, nvm

PR #4081, opened 2015-10-16
Fix issue with no tenant email domain and no invitation email

mrvisser:
I really have no idea how this got through my dev testing. I suspect I made some late changes and didn't test the case where there was no email domain for the tenant :(


mrvisser:
Assigned to @nicolaasmatthijs

PR #4083, opened 2015-10-21
Template block text for landing page sections with

mrvisser:


mrvisser:

node_modules/oae-core/tenantlandingpage/js/tenantlandingpage.js
```diff
@@ -40,6 +40,18 @@ define(['jquery', 'oae.core'], function($, oae) {
             $.ajax({
                 'url': '/api/tenant/landingPage',
                 'success': function(blocks) {
+                    // Replace all instances of ${tenantDisplayName} in block
```
nicolaasmatthijs: We have something very similar in the `i18n` api as well:

```
            if (translation !== null) {
                // Replace all of the dynamic variables, if provided
                if (variables) {
                    $.each(variables, function(dynamicVariable, dynamicTranslation) {
                        translation = translation.replace('${' + dynamicVariable + '}', dynamicTranslation);
                    });
                }
                return translation;
            }
```

I wonder if it would be worth abstracting it into a simple utility function that replaces variables in a string which can then be used in both places.

nicolaasmatthijs:
Left a suggestion. Back to @mrvisser

node_modules/oae-core/tenantlandingpage/js/tenantlandingpage.js
mrvisser: Any reason not to use trimpath for these cases?

mrvisser:
Back to @nicolaasmatthijs for feedback

node_modules/oae-core/tenantlandingpage/js/tenantlandingpage.js
nicolaasmatthijs: Would you be calling Trimpath directly or do a utility function wrapper around it?

nicolaasmatthijs:
Looks good. Merged

PR #4084, opened 2015-10-29
Deleted resource not redirecting user away

nicolaasmatthijs:
Report:

> Today I created a document, pasted in some text, disliked the formatting and so clicked the Delete button. I got the message saying the document had been deleted, but it remained on screen and I could continue to work with the text; I got error messages if I tried to share or delete it (again). When I clicked away and tried to come back it had gone.

I can't remember if we ever did redirect them away. Where would we send them? Should the back-end provide some feedback to determine which group (if any) to redirect the user to? Should we always redirect them to their activity page?


nicolaasmatthijs:
This was a regression caused by the navigational improvements in the Kingfisher release. Attached a fix. Over to @mrvisser for review.

PR #4085, opened 2015-11-02
OAE redirects to activity page when not signed in (instead of staying on the page you needed)

mrvisser:
See video for specifics: https://youtu.be/ePAc6WiazfA

When our students navigate to OAE by a link and are not signed in, they cannot post.  When they click sign in, instead of it keeping them where they wanted to post/contribute, they are redirected back to their recent activity.  In many cases, they have to retrace their steps to click the link again to get to where they needed to be.

### DIFF
--- a/node_modules/oae-core/topnavigation/js/topnavigation.js
+++ b/node_modules/oae-core/topnavigation/js/topnavigation.js
@@ -78,7 +78,6 @@ define(['jquery', 'oae.core', 'activityadapter'], function($, oae, ActivityAdapt
          */
         var initRightMenu = function() {
             if (oae.data.me.anon) {
-                console.log(JSON.stringify(oae.api.util.url(), null, 2))
                 $('#topnavigation-right', $rootel).append(oae.api.util.template().render($('#topnavigation-right-anonymous-template', $rootel), {
                     'authStrategyInfo': authStrategyInfo,
                     'redirectUrl': signInRedirectUrl || oae.api.util.url().attr('relative'),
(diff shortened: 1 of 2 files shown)


mrvisser:

node_modules/oae-core/topnavigation/js/topnavigation.js
```diff
@@ -67,7 +67,8 @@ define(['jquery', 'oae.core', 'activityadapter'], function($, oae, ActivityAdapt
         var initLeftMenu = function() {
```
nicolaasmatthijs: @mrvisser : Does this mean that this is not a bug, but a feature we never really had ?

node_modules/oae-core/topnavigation/js/topnavigation.js
```diff
@@ -67,7 +67,8 @@ define(['jquery', 'oae.core', 'activityadapter'], function($, oae, ActivityAdapt
...
             var $template = oae.data.me.anon ? $('#topnavigation-left-anonymous-template') : $('#topnavigation-left-loggedin-template');
             oae.api.util.template().render($template, {
-                'authStrategyInfo': authStrategyInfo
+                'authStrategyInfo': authStrategyInfo,
+                'redirectUrl': signInRedirectUrl || oae.api.util.url().attr('relative')
```
nicolaasmatthijs: Can we

a. Default to `oae.api.util.url().attr('relative')` inside of oae.api.authentication.getLoginRedirectUrl();`

or

b. Default to `oae.api.util.url().attr('relative')` on line 27 (`var signInRedirectUrl = oae.api.authentication.getLoginRedirectUrl();`

// Get the redirect URL
         var signInRedirectUrl =

mrvisser: We have to distinguish between there being an explicit redirect (e.g., a `?url=...` parameter) and us just "softly" indicating where to redirect after login. The former results in the signin action automatically being triggered.

node_modules/oae-core/topnavigation/js/topnavigation.js
mrvisser: It could have been a feature that was lost with between topnav refactoring and inviting guests. But I'm not entirely sure.

PR #4095, opened 2016-03-08
Configurable guest invitations bugbash follow-up

simong:
Allow you to invite guests that end up on a non-guest tenant


shared/oae/bundles/ui/default.properties
```diff
@@ -621,6 +621,9 @@ WHAT_WOULD_YOU_LIKE_TO_CALL_YOUR_GROUP = What would you like to call your group?
 WHAT_WOULD_YOU_LIKE_TO_DISCUSS = What would you like to discuss?
 WORD_DOCUMENT = Word document
 XML_DOCUMENT = XML document
+YOU_ARE_NOT_ABLE_TO_ADD_EMAIL_ADDRESSES_1 = You are not able to add ${emailAddress1}
+YOU_ARE_NOT_ABLE_TO_ADD_EMAIL_ADDRESSES_2 = You are not able to add ${emailAddress1} and ${emailAddress2}
+YOU_ARE_NOT_ABLE_TO_ADD_EMAIL_ADDRESSES_2+ = You are not able to add ${emailAddress1} and ${emailAddressesMinusOne} others
```
simong: I decided not to add in the context as that would require us to pass in the context into the oae.api.util setup autosuggest method which seemed like overkill.

nicolaasmatthijs: I would make this `Unable to add ...` and wouldn't cut this off to just 2 email addresses. I think we need to show all email addresses that could not be added explicitly in `<email1>, <email2> and <email3>` list

nicolaasmatthijs:
Finished review. Back to @simong for follow-up

nicolaasmatthijs:
Looks good. Merged.

PR #4098, opened 2016-04-18
Updated Font-Awesome to 4.6

sallakarppinen:
I included the other font files in addition to .woff as well.


nicolaasmatthijs:
A few higher level feedback items:

- As we discussed during the stand-up, PRs should be sent from a branch rather than master
- Rather than creating a completely new Pull Request, it's preferred to attach the code to the existing ticket using hub (https://github.com/github/hub)

nicolaasmatthijs:
Visual regressions:

- The icons on the landing page have increased in height by 5px. We should investigate what new FontAwesome rule is causing this and fix the regression

screen shot 2016-04-20 at 21 39 25

- The icons in the left hand navigation have moved slightly down and are no longer vertically aligned with the text

screen shot 2016-04-20 at 21 42 05

nicolaasmatthijs:

sallakarppinen:
>     As we discussed during the stand-up, PRs should be sent from a branch rather than master
>     Rather than creating a completely new Pull Request, it's preferred to attach the code to the existing ticket using hub (https://github.com/github/hub)

Acknowledged - I'll continue using this pull request for this task as it's already been created but will make sure to follow these guidelines in the future.

> The icons on the landing page have increased in height by 5px. We should investigate what new FontAwesome rule is causing this and fix the regression

I can't reproduce this - the icons seem to be one pixel wider than previously but all seem the same height (35px) to me. Which browser was this discovered with? I've tried Chromium and Firefox with no luck. Do you mean the icons themselves, not the containing element?

> The icons in the left hand navigation have moved slightly down and are no longer vertically aligned with the text

Fixed.

nicolaasmatthijs:
>> The icons on the landing page have increased in height by 5px. We should investigate what new FontAwesome rule is causing this and fix the regression
> I can't reproduce this - the icons seem to be one pixel wider than previously but all seem the same height (35px) to me. Which browser was this discovered with? I've tried Chromium and Firefox with no luck. Do you mean the icons themselves, not the containing element?

I'm using Firefox 45.0.2.

Before:

screen shot 2016-04-21 at 07 24 47

screen shot 2016-04-21 at 07 25 26

After:

screen shot 2016-04-21 at 07 22 11

screen shot 2016-04-21 at 07 25 51

sallakarppinen:
Ah, sorry. I think that should fix it.

sallakarppinen:
A question regarding the higher level feedback from earlier: I tried hub and using the issue-to-pull request feature it offers and got a warning saying 'Issue to pull request conversion is deprecated and might not work in the future.' Should I keep using it and just make sure to never update the app or is there another method to achieve the same?

mrvisser:
@sallakarppinen yea feel free to ignore the warning until the feature is removed, because for our workflow it keeps things a bit cleaner. For us, it's rare to have an issue opened, and have multiple substantially different approaches come in as PRs as we iterate over the issue. The problem of separate issue from PR for us is that it tends to split the conversation and leave cruft in the issue tracker after a PR is merged.

When the endpoint is removed we'll likely just iterate over this process, but I wouldn't be surprised if the endpoint is removed with a suitable alternative for this kind of use-case (i.e., better issue-linking or something).

sallakarppinen:
Closed in favour of #4128.

PR #4099, opened 2016-04-28
Users with incorrect display name get tenant visibility when editing profile

sallakarppinen:


nicolaasmatthijs:
@sallakarppinen : It looks like the FontAwesome work is part of this PR as well. Would you be able to submit a PR from a branch that doesn't have this included so we can keep the PRs focussed?

sallakarppinen:
@nicolaasmatthijs, done - seems that I can't close pull requests here so could you close this one for me?

PR #4100, opened 2016-04-28
Users with incorrect display name get tenant visibility when editing profile

sallakarppinen:


node_modules/oae-core/editprofile/js/editprofile.js
```diff
@@ -101,8 +101,12 @@ define(['jquery', 'oae.core'], function ($, oae) {
         var renderEditProfileMainPanel = function() {
             // If the display name is not valid, clear it to inform the template that the user
             // has no real display name
+            var tenantVisibility = oae.api.config.getValue('oae-principals', 'user', 'visibility');
             if (!isValidDisplayName()) {
                 profile.displayName = null;
+                if(profile.visibility === 'private' && tenantVisibility !== 'private') {
```
nicolaasmatthijs: Needs a space after `if` to comply with the styleguide

nicolaasmatthijs: But it might actually be easier to just always do `profile.visibility = tenantVisibility;` when the display name is not valid.

nicolaasmatthijs:

node_modules/oae-core/editprofile/js/editprofile.js
```diff
@@ -101,8 +101,10 @@ define(['jquery', 'oae.core'], function ($, oae) {
         var renderEditProfileMainPanel = function() {
             // If the display name is not valid, clear it to inform the template that the user
             // has no real display name
+            var tenantVisibility = oae.api.config.getValue('oae-principals', 'user', 'visibility');
             if (!isValidDisplayName()) {
                 profile.displayName = null;
+                profile.visibility = tenantVisibility;
```
nicolaasmatthijs: Can you just add a line of docs (and a reference to this ticket) to this ?

Note: the format for these docs would be:

```
// Comment
// @see https://github.com/oaeproject/3akai-ux/pull/4100
```

nicolaasmatthijs:
Added final comment. Back to @sallakarppinen for follow-up

node_modules/oae-core/editprofile/js/editprofile.js
```diff
@@ -103,6 +103,11 @@ define(['jquery', 'oae.core'], function ($, oae) {
             // has no real display name
             if (!isValidDisplayName()) {
```
nicolaasmatthijs: The `isValidDisplayName` validation function should have exactly the same checks as the back-end checks. It looks like it currently checks for `http(s)` and `@`, but not `shibboleth!`. Can you add that in as well?

nicolaasmatthijs:

sallakarppinen:
I've added shibboleth check to the front end (also put all the tests into single regex, same as is being used in the backend).

nicolaasmatthijs:
Looks good. Merged!

PR #4102, opened 2016-05-05
Tile margin issue

nicolaasmatthijs:
Reported by @D7Torres:

As you can see, the padding of the icon of the tile is overflowing the tile, because the margin is negative. My problem was, if you try to click in the button you could click the tile instead. You could even click a tile of the first road and actually being clicking the tile below, in the second row.

They only thing I had to do to solve it was changing this styles in `oae.components.css`

```
.oae-tile .oae-thumbnail.oae-thumbnail-large:before {
    margin-top: -48px;
    padding-top: 80px;
}
```

to 

```
.oae-tile .oae-thumbnail.oae-thumbnail-large:before {
    margin-top: -0px;
    padding-top: 32px;
 }
```

image1
image2

### DIFF
--- a/node_modules/oae-core/activity/css/activity.css
+++ b/node_modules/oae-core/activity/css/activity.css
@@ -84,11 +84,6 @@
     height: 225px;
 }
 
-/* Re-position the standard resourceType icon because of the changed tile height */
-.activity-widget .activity-preview-container ul.oae-list > li .oae-thumbnail.oae-thumbnail-large:before {
-    margin-top: -43px;
-}
-
 .activity-widget .activity-show-all-toggle-container {
     margin-bottom: -15px;
     margin-top: 20px;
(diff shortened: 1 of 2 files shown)


nicolaasmatthijs:
Assigning to @sallakarppinen for review.

@sallakarppinen : Next to reviewing the code, you should also do a full functional review of the fix. In this case, that would mean going through the different places in which list tiles are being used and ensuring that no regressions have happened. If everything looks good to go, feel free to merge.

sallakarppinen:
I realise it's probably unrelated, but the Travis build seems to fail quite often (as it has here). How much attention should I pay to it?

nicolaasmatthijs:
@sallakarppinen : Feel free to ignore the Travis build on the 3akai-ux side. There's a number of fixes that we need to put in place.

PR #4089, opened 2016-05-06
Folder library not requesting more items when `nextToken` is specified

sallakarppinen:
When the UI requests a folder library, if the results array is empty, it will stop requesting regardless if there is a `nextToken` defined by the back-end.

It should continue requesting items until the `nextToken` is falsey.


shared/oae/js/jquery-plugins/jquery.infinitescroll.js
```diff
@@ -165,7 +165,7 @@ define(['jquery', 'underscore', 'oae.api.util', 'oae.api.i18n', 'oae.api.l10n'],
          * the postProcessor as well.
          *
          * @param  {Object}      data           List of items to add to the infinite scroll list
-         * @param  {Objet}      [opts]          Optional arguments
+         * @param  {Object}     [opts]          Optional arguments
```
nicolaasmatthijs: Good catch :)

nicolaasmatthijs:
Looks good. Merged.

PR #4105, opened 2016-05-26
Group description (About) and discussion description should both support markdown

sallakarppinen:
We currently render markdown in comments (I think it's client-side), we should add that handling in these areas as well to make it more consistent.

### DIFF
--- a/shared/oae/api/oae.api.util.js
+++ b/shared/oae/api/oae.api.util.js
@@ -1656,12 +1656,16 @@ define(['exports', 'require', 'jquery', 'underscore', 'oae.api.config', 'markdow
                 return '';
             } else {
 
+                // Escape asterisks instead of treating them as italics markers
+                input = input.replace(/\*/g, "\\*").replace(/\\\\/g, "\\");
+
                 // Convert the Markdown input string to HTML using marked.js. `gfm`
                 // automatically recognizes text beginning with http: or https: as a URL
                 // and converts it to a link. We also specify that the input should be sanitized.
                 // @see https://github.com/chjj/marked
                 input = markdown(input.toString(), {
                     'gfm': true,
+                    'tables': true,
                     'breaks': true,
                     'sanitize': true
                 });
--- a/shared/oae/css/oae.base.css
+++ b/shared/oae/css/oae.base.css
@@ -903,24 +903,23 @@ ul.as-list li.as-result-item.active {
  */
 
 .oae-markdown-embedded table {
-    margin: 20px 0;
     border-bottom: 1px solid #ECECEC;
-    color: #333;
     font-size: 14px;
+    margin: 20px 0;
 }
 
 .oae-markdown-embedded table th {
-    background-color: #FFF;
-    border-top: 1px solid #ECECEC;
+    background-color: transparent;
     border-bottom: 1px solid #999;
-    padding: 10px 0;
+    border-top: 1px solid #ECECEC;
     font-weight: bolder;
+    padding: 10px 0;
 }
 
 .oae-markdown-embedded table td {
-    padding: 5px 20px 6px 0;
     font-weight: normal;
     height: 22px;
+    padding: 6px 20px 6px 0;
 }
 
 .oae-markdown-embedded table tr:nth-child(odd) {
--- a/shared/oae/css/oae.base.css
+++ b/shared/oae/css/oae.base.css
@@ -909,11 +909,11 @@ ul.as-list li.as-result-item.active {
 }
 
 .oae-markdown-embedded table th {
-    background-color: transparent;
+    background-color: #FFF;
     border-bottom: 1px solid #999;
     border-top: 1px solid #ECECEC;
     font-weight: bolder;
-    padding: 10px 0;
+    padding: 10px;
 }
 
 .oae-markdown-embedded table td {
@@ -922,6 +922,6 @@ ul.as-list li.as-result-item.active {
     padding: 6px 20px 6px 0;
 }
 
-.oae-markdown-embedded table tr:nth-child(odd) {
+.oae-markdown-embedded table tr:nth-child(even) {
     background-color: #ECECEC;
 }
--- a/shared/oae/css/oae.base.css
+++ b/shared/oae/css/oae.base.css
@@ -904,7 +904,6 @@ ul.as-list li.as-result-item.active {
 
 .oae-markdown-embedded table {
     border-bottom: 1px solid #ECECEC;
-    font-size: 14px;
     margin: 20px 0;
 }
 
@@ -912,16 +911,20 @@ ul.as-list li.as-result-item.active {
     background-color: #FFF;
     border-bottom: 1px solid #999;
     border-top: 1px solid #ECECEC;
-    font-weight: bolder;
+    font-weight: 500;
     padding: 10px;
 }
 
 .oae-markdown-embedded table td {
     font-weight: normal;
     height: 22px;
-    padding: 6px 20px 6px 0;
+    padding: 6px 10px;
 }
 
 .oae-markdown-embedded table tr:nth-child(even) {
     background-color: #ECECEC;
+}
+
+.oae-markdown-embedded table tr:nth-child(odd) {
+    background-color: #FFF;
 }
--- a/shared/oae/css/oae.base.css
+++ b/shared/oae/css/oae.base.css
@@ -897,3 +897,32 @@ ul.as-list li.as-result-item.active {
 .oae-markdown-embedded p:first-child {
     margin-top: 0;
 }
+
+/**
+ * Markdown table styles
+ */
+
+.oae-markdown-embedded table {
+    margin: 20px 0;
+    border-bottom: 1px solid #ECECEC;
+    color: #333;
+    font-size: 14px;
+}
+
+.oae-markdown-embedded table th {
+    background-color: #FFF;
+    border-top: 1px solid #ECECEC;
+    border-bottom: 1px solid #999;
+    padding: 10px 0;
+    font-weight: bolder;
+}
+
+.oae-markdown-embedded table td {
+    padding: 5px 20px 6px 0;
+    font-weight: normal;
+    height: 22px;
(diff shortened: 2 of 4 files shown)


nicolaasmatthijs:
@sallakarppinen : It looks like this includes the FontAwesome branch as well. Can you exclude that from this PR?

nicolaasmatthijs:
@sallakarppinen : As part of this PR, can you exclude `*` processing from our MarkDown processor so we can avoid having to escape things such as `*Research` and `*Unity` ?

nicolaasmatthijs:

mrvisser:
> As part of this PR, can you exclude * processing from our MarkDown processor so we can avoid having to escape things such as *Research and *Unity ?

Do we do this for comments as well?

nicolaasmatthijs:
> Do we do this for comments as well?

Not yet, but we should do it for comments as well.

mrvisser:
That should probably be done as part of this ticket as well then.

nicolaasmatthijs:
Yes, absolutely.

mrvisser:
And I think that's what you meant :P

sallakarppinen:
Can I also add support for markdown tables?

nicolaasmatthijs:
@sallakarppinen : Can you squash these commits so adding and removing FontAwesome isn't part of the commit history?

shared/oae/api/oae.api.util.js
```diff
@@ -1656,12 +1656,15 @@ define(['exports', 'require', 'jquery', 'underscore', 'oae.api.config', 'markdow
                 return '';
             } else {
 
+                input = input.replace(/\*/g, "\\*");
```
nicolaasmatthijs: Can you add a line of documentation here?

nicolaasmatthijs: We also have a number of comments in production that already escape `*` to avoid the italic behavior. Can we make sure that `*`s that are already escaped are not further escaped?

shared/oae/api/oae.api.util.js
```diff
@@ -1656,12 +1656,15 @@ define(['exports', 'require', 'jquery', 'underscore', 'oae.api.config', 'markdow
...
+
                 // Convert the Markdown input string to HTML using marked.js. `gfm`
                 // automatically recognizes text beginning with http: or https: as a URL
                 // and converts it to a link. We also specify that the input should be sanitized.
                 // @see https://github.com/chjj/marked
                 input = markdown(input.toString(), {
                     'gfm': true,
+                    'tables': true,
```
nicolaasmatthijs: @SamPeck : Can you provide a design for how you want tables inside of a comment to be rendered (margin, padding, border style, etc.)?

screen shot 2016-05-27 at 12 10 39

nicolaasmatthijs:
Assigning to @SamPeck for design feedback

sallakarppinen:
Have made the changes suggested.

shared/oae/api/oae.api.util.js
SamPeck: [tables.pdf](https://github.com/oaeproject/3akai-ux/files/291482/tables.pdf)
tables

nicolaasmatthijs:
Back to @sallakarppinen for implementation of table styling

shared/oae/api/oae.api.util.js
nicolaasmatthijs: @SamPeck : Are the table colors skin-independent or should they use certain skin values?

shared/oae/css/oae.base.css
```diff
@@ -897,3 +897,32 @@ ul.as-list li.as-result-item.active {
...
+
+.oae-markdown-embedded table td {
+    padding: 5px 20px 6px 0;
+    font-weight: normal;
+    height: 22px;
+}
+
+.oae-markdown-embedded table tr:nth-child(odd) {
+    background-color: #ECECEC;
+}
```
nicolaasmatthijs: Missing newline

shared/oae/css/oae.base.css
```diff
@@ -897,3 +897,32 @@ ul.as-list li.as-result-item.active {
...
 }
+
+/**
+ * Markdown table styles
+ */
+
+.oae-markdown-embedded table {
+    margin: 20px 0;
+    border-bottom: 1px solid #ECECEC;
+    color: #333;
```
nicolaasmatthijs: The text color is already set in the skin (and is set to `#333` by default)

shared/oae/css/oae.base.css
```diff
@@ -897,3 +897,32 @@ ul.as-list li.as-result-item.active {
...
     margin-top: 0;
 }
+
+/**
+ * Markdown table styles
+ */
+
+.oae-markdown-embedded table {
+    margin: 20px 0;
+    border-bottom: 1px solid #ECECEC;
```
nicolaasmatthijs: We try to avoid defining colors in the base CSS file because of skinning reasons. I've left a question for Sam to check if these should be hardcoded or driven by skin config values.

shared/oae/css/oae.base.css
```diff
@@ -897,3 +897,32 @@ ul.as-list li.as-result-item.active {
...
+    color: #333;
+    font-size: 14px;
+}
+
+.oae-markdown-embedded table th {
+    background-color: #FFF;
```
nicolaasmatthijs: `transparent` might be better here

shared/oae/css/oae.base.css
```diff
@@ -897,3 +897,32 @@ ul.as-list li.as-result-item.active {
...
+    border-top: 1px solid #ECECEC;
```
nicolaasmatthijs: Alphabetical sorting

shared/oae/css/oae.base.css
```diff
@@ -897,3 +897,32 @@ ul.as-list li.as-result-item.active {
...
+    border-bottom: 1px solid #999;
+    padding: 10px 0;
+    font-weight: bolder;
+}
+
+.oae-markdown-embedded table td {
+    padding: 5px 20px 6px 0;
```
nicolaasmatthijs: Padding top and bottom should probably be the same (6px)

nicolaasmatthijs:
Finished review and left question for @SamPeck

sallakarppinen:
Made changes based on comments except to those waiting for feedback from Sam.

shared/oae/api/oae.api.util.js
SamPeck: The only skinnable element should be the text colour using "Plain text color" value, the lines and BG striping seems a bit overkill to ask administrators to configure, or assigning values to arbitrary styles could be confusing and problematic.

shared/oae/api/oae.api.util.js
```diff
@@ -1656,12 +1656,16 @@ define(['exports', 'require', 'jquery', 'underscore', 'oae.api.config', 'markdow
                 return '';
             } else {
 
+                // Escape asterisks instead of treating them as italics markers
+                input = input.replace(/\*/g, "\\*").replace(/\\\\/g, "\\");
```
nicolaasmatthijs: Wouldn't it be easier to just replace 'not a backslash followed by a *' ?

shared/oae/css/oae.base.css
```diff
@@ -897,3 +897,31 @@ ul.as-list li.as-result-item.active {
 .oae-markdown-embedded p:first-child {
     margin-top: 0;
 }
+
+/**
+ * Markdown table styles
```
nicolaasmatthijs: There's still some styling issues:

screen shot 2016-06-02 at 08 15 55

nicolaasmatthijs: - The header row has a grey background, so the `transparent` property isn't working
- There's no padding enforced between the table headers
- The first row under the table header should be white, only the one after that should be grey

nicolaasmatthijs:

shared/oae/api/oae.api.util.js
sallakarppinen: I think this is _easier_, since that would require a negative lookbehind, which doesn't exist in JS?

shared/oae/css/oae.base.css
sallakarppinen: Should now be fixed.

nicolaasmatthijs: This still looks a bit off:

screen shot 2016-06-02 at 11 28 05

nicolaasmatthijs: - I'm not convinced that the font size of tables should be larger than the regular font size, so let's remove that
- The left and right padding on regular table cells should be the same as those on table header cells. Otherwise they don't align
- We try not to use `bolder`. We use either `bold` or 500/600/700
- We probably do want to give the odd rows an explicit white background color. Currently it looks a bit odd in the activity feed because of the grey background.

nicolaasmatthijs:
Back to @sallakarppinen

sallakarppinen:
Done.

shared/oae/css/oae.base.css
```diff
@@ -904,24 +904,27 @@ ul.as-list li.as-result-item.active {
...
-    font-size: 14px;
     margin: 20px 0;
 }
 
 .oae-markdown-embedded table th {
     background-color: #FFF;
     border-bottom: 1px solid #999;
     border-top: 1px solid #ECECEC;
-    font-weight: bolder;
+    font-weight: 500;
```
nicolaasmatthijs: 500 is not showing as bold to me. This should be 600 or 700 to match the style in the designs.

sallakarppinen: Me neither, but it is what @SamPeck specified above. Can swap to 600 though.

nicolaasmatthijs: Yeah, I noticed, but it needs to visually match the design. I've changed it to 600 and will be merging this.

sallakarppinen: Cool, thanks. Sorry about that.

nicolaasmatthijs:
Followed up and merged.

PR #4106, opened 2016-05-27
Handle MediaCore failures more gracefully

sallakarppinen:
If the REST endpoint at `/api/mediacore/embed/:contentId` fails (maybe mediacore went down, maybe there's something our end, ..) the UI doesn't display anything. It should probably show a message and offer the download button.


sallakarppinen:
Ok, I've removed MediaCore functionality from 3akai-ux. (Although as a side, it seems to still work in some places - I could preview [this video](https://research.unity.ac/content/gatech/gymRz1_rkl) without issues.) Uploaded videos will now have no preview but just a download button. I'm assuming we will want to replace the functionality with something else in the future, although it doesn't look like a very widely used feature?

nicolaasmatthijs:
We were told that the MediaCore service would cease to exist, but it does look like it is still available. Therefore, we shouldn't remove the integration yet, and we should probably focus on generating an alternative. I'm going to close this ticket in favour of a replacement ticket.

Issue #4108, opened 2016-06-07
Discussion / Group description markdown has too much spacing in lists


mrvisser:
Sample content:

```
This is my comment with a list:

1. Ok there's this
2. then this

How is that specing?
```

Appearance discrepancy between discussion profile and comment:

![image](https://cloud.githubusercontent.com/assets/102265/15861356/2935712c-2c9a-11e6-96a3-c1eb3682f6dc.png)

I feel like the comment spacing is desired.

My browser is Google Chrome 50.0.2661.102 (64-bit). Looking at the inspector, I can't actually figure out where the extra spacing is coming from. It appears just blank white space unaccounted for by any margin or padding.

nicolaasmatthijs:
Yes, the comments spacing is desired. Assigning to @sallakarppinen for fix

sallakarppinen:
I've included the fix for this in the pull request #4110.

PR #4110, opened 2016-06-08
Markdown tables should use alternative CSS in description

sallakarppinen:
From bugbash: tables in comments display very different than in description. @SamPeck supplied alternative values for the main container (see below).

tablecontentcontaineralt-v01_960


sallakarppinen:
I've also included the fix for #4108 here to avoid merge conflicts if they were done separately.

PR #4109, opened 2016-06-16
Omitting * from markdown

sallakarppinen:
I know this was a conscious decision, here's a bit of discussion in slack about the change during the bug bash:

```

mrvisser [10:35 AM] 
no it’s a good question, I posted above about it, what do you think?

[10:35] 
> Should we support an `*` followed by a space to be a bulleted list still?
> I personally use bulleted list a lot when I’m markdowning
> and it doesn’t affect *Research

cnicoletti [10:35 AM] 
I am one who doesnt use markdown often, but that is like THE one I use

mrvisser [10:35 AM] 
the bulleted list?

cnicoletti [10:36 AM] 
Ahh, I see now, yes, I would say so

mrvisser [10:36 AM] 
or the emphasis? or both?

cnicoletti [10:36 AM] 
both

[10:36] 
but definitely for bulleting

[10:36] 
but I see how I can just use - lists

mrvisser [10:37 AM] 
I didn’t know that
```

`*` I think is a pretty important character to have. Is there a less intrusive way we can have an exception for `*Research`?

At the least, maybe a little legend at the bottom of textareas that support markdown. For example, I've seen a simple list of supported syntax on a single line at the bottom of a text area like:

`*bold*, _italic_, * bulleted list, -strikethrough-`

### DIFF
--- a/shared/oae/api/oae.api.util.js
+++ b/shared/oae/api/oae.api.util.js
@@ -1657,7 +1657,7 @@ define(['exports', 'require', 'jquery', 'underscore', 'oae.api.config', 'markdow
             } else {
 
                 // Escape asterisks instead of treating them as italics markers
-                input = input.replace(/(\*)(?=[RU])/g, "\\*").replace(/(?:\s)(\*)(?=\s)/g, " \\*").replace(/\\\\/g, "\\");
+                input = input.replace(/(\*)(?=Research|Unity)/g, "\\*").replace(/(?:\s)(\*)(?=\s)/g, " \\*").replace(/\\\\/g, "\\");
 
                 // Convert the Markdown input string to HTML using marked.js. `gfm`
                 // automatically recognizes text beginning with http: or https: as a URL
--- a/shared/oae/api/oae.api.util.js
+++ b/shared/oae/api/oae.api.util.js
@@ -1655,9 +1655,29 @@ define(['exports', 'require', 'jquery', 'underscore', 'oae.api.config', 'markdow
             if (!input) {
                 return '';
             } else {
-
+                var nonWord = /\W/;
                 // Escape asterisks instead of treating them as italics markers
-                input = input.replace(/(\*)(?=Research|Unity)/g, "\\*").replace(/(?:\s)(\*)(?=\s)/g, " \\*").replace(/\\\\/g, "\\");
+                input = input.replace(/\*/g, function(match, index) {
+                    var prevChar = input.charAt(index - 1);
+
+                    // Is this an asterisk used at the start of a line to indicate a bulleted list?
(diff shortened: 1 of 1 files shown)


sallakarppinen:
I don't think I can get the exact same functionality as Github without lookbehinds, but I've set it to only escape asterisks when a) at the start of the word if followed by capital R or U, or b) in the middle of a string but surrounded by whitespace. I *think* that using gfm option it'll also ignore asterisks when there's no closing asterisk on the same line, but not certain. Seems to work that way on my local.

nicolaasmatthijs:
@sallakarppinen : Can we implement the emphasising using the following rule:

- Emphasise everything between a `*` at the beginning of a word followed by a `*` as the last character of a word.
- Escape all other `*`s

nicolaasmatthijs:

sallakarppinen:
@nicolaasmatthijs: Do we want to be able to do bulleted lists with asterisks still?

shared/oae/api/oae.api.util.js
```diff
@@ -1655,9 +1655,29 @@ define(['exports', 'require', 'jquery', 'underscore', 'oae.api.config', 'markdow
             if (!input) {
                 return '';
             } else {
-
+                var nonWord = /\W/;
                 // Escape asterisks instead of treating them as italics markers
-                input = input.replace(/\*/g, "\\*").replace(/\\\\/g, "\\");
+                input = input.replace(/\*/g, function(match, index) {
```
nicolaasmatthijs: I think the rules we want to follow can be captured in a regex:

For every line:

- all `*`s at the beginning of the line or `*`s preceded by a space
- and followed by a character different than a space
- followed by a different `*` preceded by a character different than a space
- and followed by a space or the end of the line

For every such combination of `*`s, the content should be made italic. All other `*`s should be escaped.

nicolaasmatthijs:
Back to @sallakarppinen

shared/oae/api/oae.api.util.js
sallakarppinen: I don't think they can, at least not nicely, as javascript doesn't allow for lookbehind in regex (so there isn't a way to check the preceding character). We would need to check every space and letter character to see if they are followed by an asterisk, then check the character following that asterisk - I think doing it this way is quicker.

sallakarppinen:
@nicolaasmatthijs, IIRC you were looking into finding a regex based solution for this, as I had been unable to craft one that doesn't use lookbehinds. Did you get anywhere with that or should we go with what we have now?

sallakarppinen:
Looks like this can actually be fixed by swapping our markdown parser to showdown...

PR #4111, opened 2016-07-11
The "this tenant only" checkbox in the UI should be hidden for private tenants

sallakarppinen:
Related to [Hilary issue 1295](https://github.com/oaeproject/Hilary/issues/1295).

### DIFF
--- a/ui/js/search.js
+++ b/ui/js/search.js
@@ -101,15 +101,16 @@ require(['jquery', 'oae.core', 'underscore', 'jquery.history', 'jquery.switchtab
             $('#search-refine-type input[type="checkbox"][data-type="' + type + '"]').prop('checked', true);
         });
 
-        // Ensure only the selected tenant is chosen, if at all
-        $('#search-refine-tenant input[type="checkbox"]').removeAttr('checked');
-        if (queryData.tenant) {
-            $('#search-refine-tenant input[type="checkbox"][data-tenant="' + queryData.tenant + '"]').prop('checked', true);
-        }
-
         // Hide the 'within tenant' checkbox for private tenants
-        if (oae.api.config.getValue('oae-tenants', 'tenantprivacy', 'tenantprivate')) {
+        var isPrivate = oae.api.config.getValue('oae-tenants', 'tenantprivacy', 'tenantprivate');
+        if (isPrivate) {
             $('#search-refine-tenant').hide();
+        } else {
+            // Ensure only the selected tenant is chosen, if at all
+            $('#search-refine-tenant input[type="checkbox"]').removeAttr('checked');
+            if (queryData.tenant) {
+                $('#search-refine-tenant input[type="checkbox"][data-tenant="' + queryData.tenant + '"]').prop('checked', true);
+            } 
         }
 
         var searchParams = {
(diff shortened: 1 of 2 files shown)


sallakarppinen:
Should now also fix the issue where private tenancies were showing results from outside the tenancy by forcing search scope for them.

ui/js/search.js
```diff
@@ -101,13 +101,18 @@ require(['jquery', 'oae.core', 'underscore', 'jquery.history', 'jquery.switchtab
...
         });
 
-        // Ensure only the selected tenant is chosen, if at all
-        $('#search-refine-tenant input[type="checkbox"]').removeAttr('checked');
-        if (queryData.tenant) {
-            $('#search-refine-tenant input[type="checkbox"][data-tenant="' + queryData.tenant + '"]').prop('checked', true);
+        // Hide the 'within tenant' checkbox for private tenants
+        var isPrivate = oae.api.config.getValue('oae-tenants', 'tenantprivacy', 'tenantprivate');
+        if (isPrivate) {
+            $('#search-refine-tenant').hide();
```
nicolaasmatthijs: The checkbox should be hidden by default and shown when the tenant is not private to prevent flickering on private tenants.

ui/js/search.js
```diff
@@ -101,13 +101,18 @@ require(['jquery', 'oae.core', 'underscore', 'jquery.history', 'jquery.switchtab
...
+        } else {
+            // Ensure only the selected tenant is chosen, if at all
```
nicolaasmatthijs: Do we need to programatically check/uncheck this box at all? Wouldn't this be taken care off by the back-end feed?

ui/js/search.js
```diff
@@ -119,8 +124,9 @@ require(['jquery', 'oae.core', 'underscore', 'jquery.history', 'jquery.switchtab
             // "My" search never includes users, and will be harmful if only "user" is selected from
             // the "Everything" tab. Just always filter it out
             searchParams.resourceTypes = _.without(searchParams.resourceTypes, 'user');
-        } else if (searchParams.scope === '_all' && queryData.tenant) {
-            // If a tenant is specified, we search by the tenant
+        } else if (searchParams.scope === '_all' && (queryData.tenant || isPrivate)) {
+            // If a tenant is specified or the current tenant is private, we search by the tenant
```
nicolaasmatthijs: Same here. Wouldn't this be taken care off by the back-end feed?

nicolaasmatthijs: If we do need to specify the tenant id, it should be taken from the `oae.data.me` object, rather than extracting it from the HTML

nicolaasmatthijs:
Finished review. Back to @sallakarppinen for follow-up.

ui/js/search.js
sallakarppinen: I don't think it would be since the scope can be changed by changing the tab. Noted on `oae.data.me` though.

ui/js/search.js
sallakarppinen: I'm not sure. I imagine it could've been set up this way for the case where a search is limited to a tenancy other than the user's own?

mrvisser: I don't think the back-end does anything fancy with assuming whether or not the UI will want to show items for the current tenant or the whole system.

The back-end is designed to just ensure the privacy/security constraints, and then what the UI wants to project to the user is more of a front-end experience detail. So it's important for the UI to say explicitly "this tenant only", "only for that tenant", etc... explicitly.

PR #4112, opened 2016-08-01
Mediacore replacement

sallakarppinen:


node_modules/oae-core/filepreview/filepreview.html
```diff
@@ -54,6 +54,32 @@
     </div>
 --></div>
```
nicolaasmatthijs: The mediacore preview can be removed

node_modules/oae-core/filepreview/filepreview.html
```diff
@@ -54,6 +54,32 @@
...
+            {var mime = content.mime}
+        {/if}
+
+        {if content.previews && content.previews.height} 
+            {var height = content.previews.height}
+        {else}
+            {var height = '1000px'}
+        {/if}
+
+        <div>
```
nicolaasmatthijs: Do we need another div within the `filepreview-video-container` div ?

node_modules/oae-core/filepreview/filepreview.html
```diff
@@ -54,6 +54,32 @@
...
 --></div>
 
+<!-- VIDEO PREVIEW -->
+<div id="filepreview-video-template"><!--
+    <div id="filepreview-video-container" class="text-center">
+        {if content.previews && content.previews.video}
+            {var url = content.previews.video.replace('remote:', '')}
+            {var mime = 'video/mp4'}
+        {else}
+            {var url = content.downloadPath}
```
nicolaasmatthijs: In what scenario do we want to show a video player without there being a generated preview for it ?

node_modules/oae-core/filepreview/filepreview.html
```diff
@@ -54,6 +54,32 @@
...
+            {var mime = content.mime}
+        {/if}
+
+        {if content.previews && content.previews.height} 
+            {var height = content.previews.height}
```
nicolaasmatthijs: The height of the player should just follow the rules we have set up for YouTube videos (i.e., 600px on larger screens and 400px on mobile devices)

node_modules/oae-core/filepreview/js/filepreview.js
```diff
@@ -80,6 +80,20 @@ define(['require', 'jquery', 'oae.core'], function(require, $, oae) {
...
+        /////////////////////
+        //  VIDEO PREVIEW  //
+        /////////////////////
+
+        /**
+         * Render a preview of a video file
+         */
+        var renderVideoPreview = function() {
+            oae.api.util.template().render($('#filepreview-video-template', $rootel), {'content': widgetData}, $('#filepreview-container', $rootel));
+            videojs('filepreview-video-player', {'controls': true, 'autoplay': false, 'preload': 'auto', 'poster': widgetData.previews.mediumUrl}, function() {
```
nicolaasmatthijs: No need to pass in an empty callback

node_modules/oae-core/filepreview/js/filepreview.js
```diff
@@ -80,6 +80,20 @@ define(['require', 'jquery', 'oae.core'], function(require, $, oae) {
...
 
 
+        /////////////////////
+        //  VIDEO PREVIEW  //
+        /////////////////////
+
+        /**
+         * Render a preview of a video file
+         */
+        var renderVideoPreview = function() {
```
nicolaasmatthijs: When ZenCoder is processing the video, we see the `Processing` message in the UI. However, when processing has finished, the UI should automatically be showing the preview following a push notification. It's doing this for other previews, but doesn't appear to be doing this for videos.

shared/oae/css/oae.core.css
```diff
@@ -39,3 +39,6 @@
 /* Import the OAE Component Styles. These contain OAE specific re-usable styles that,
    combined with basic bootstrap, should make it easy to quickly put new widgets together */
 @import url('../../oae/css/oae.components.css');
+
+/* Import the Video.js CSS file */
```
nicolaasmatthijs: Can we import this above the OAE styles ?

nicolaasmatthijs:

node_modules/oae-core/filepreview/filepreview.html
sallakarppinen: Cool, I'll take all the other mediacore code out as well then.

node_modules/oae-core/filepreview/filepreview.html
sallakarppinen: The idea was that if for some reason Zencoder encoding failed, if the video happened to be in one of the HTML5/Flash supported formats we could still play it. I could flesh this out a bit to check the mime-type and whether it is one of them, but I thought it would be a good fallback even if other formats are not as widely supported as mp4.

node_modules/oae-core/filepreview/filepreview.html
sallakarppinen: Cool - I'll also change the Zencoder settings in the Hilary pr to limit the height of video to a max of 600px (no point making it bigger if we won't display the excess).

node_modules/oae-core/filepreview/js/filepreview.js
sallakarppinen: I think this will be something that's happening on the Hilary side, will look into it.

node_modules/oae-core/filepreview/filepreview.html
nicolaasmatthijs: This is something we tried in the past, and the results were very inconsistent across browsers/user machines so it might be safest to not show the preview unless we were able to generate one through ZenCoder.

node_modules/oae-core/filepreview/filepreview.html
nicolaasmatthijs: It looks like the video player itself has absolute positioning. Overriding `.video-js` with `position: unset` appears to make it use the container's dimensions.

sallakarppinen:
Thanks for figuring out the CSS issue, @nicolaasmatthijs. I've made some fixes and deployed them to qa2 server.

node_modules/oae-core/filepreview/filepreview.html
```diff
@@ -47,35 +47,20 @@
...
-    </div>
---></div>
-
 <!-- VIDEO PREVIEW -->
 <div id="filepreview-video-template"><!--
     <div id="filepreview-video-container" class="text-center">
-        {if content.previews && content.previews.video}
-            {var url = content.previews.video.replace('remote:', '')}
-            {var mime = 'video/mp4'}
+        {if !isAudio}
```
nicolaasmatthijs: It'd be more readable to do the following:

```
 {if isAudio}
   {var tag = 'audio'}
{else}
  {var tag = 'video'}
{/if}
```

shared/vendor/css/videojs/video-js.css
```diff
@@ -236,7 +236,7 @@
   box-sizing: border-box;
   color: #fff;
   background-color: #fff;
-  position: relative;
+  position: unset;
```
nicolaasmatthijs: Can we override this in the OAE CSS rather than change the vendor code? This will ensure that updating video.js will more likely not have any regressions.

node_modules/oae-core/filepreview/css/filepreview.css
```diff
@@ -58,39 +58,27 @@
...
-}
-
-.mobile #filepreview-container #filepreview-mediacore iframe {
-    width: 340px; /* iOS will need an explicit width for the iframe */
-}
-
-#filepreview-container #filepreview-mediacore.filepreview-mediacore-audio iframe {
-    height: 24px;
-}
+/* Video preview */
```
nicolaasmatthijs: Do you know if there's a way to make `video.js` fill the available player space, similar to how YouTube embedding works?

screen shot 2016-08-01 at 11 34 55

node_modules/oae-core/filepreview/css/filepreview.css
```diff
@@ -58,39 +58,27 @@
...
 
 @media (min-width: 780px) {
```
nicolaasmatthijs: The boundaries of the player should align with the boundaries of the comments widget.

node_modules/oae-core/filepreview/js/filepreview.js
nicolaasmatthijs: This is probably expected, but this is still a problem on QA2

nicolaasmatthijs:
Finished another round of review. Back to @sallakarppinen

node_modules/oae-core/filepreview/css/filepreview.css
sallakarppinen: It actually does that by default (the background is black), I'll change it back.

node_modules/oae-core/filepreview/css/filepreview.css
sallakarppinen: The other changes mentioned here are now done, but I'm a bit unclear on what you mean by this one?

nicolaasmatthijs: In the screenshot above, you can see that the video controls exceed the edges of the comments widget on the left and the right. The left side of the video player should be aligned with the left edge of the user's profile picture and the right side of the video player should be aligned with the right edge of the comment box.

node_modules/oae-core/filepreview/css/filepreview.css
```diff
@@ -82,3 +82,12 @@
...
 }
+
+.video-js {
+    position: unset;
+    background-color: #fff;
+}
+
+.vjs-tech {
+    background-color: #000;
+}
```
nicolaasmatthijs: Needs newline

node_modules/oae-core/filepreview/css/filepreview.css
```diff
@@ -82,3 +82,12 @@
         height: 200px;
     }
 }
+
```
nicolaasmatthijs: Can you add a note here that indicate that these are `video.js` overrides and why they're necessary ?

node_modules/oae-core/filepreview/css/filepreview.css
```diff
@@ -90,4 +90,11 @@
 
 .vjs-tech {
     background-color: #000;
+    width: 98% !important;
```
nicolaasmatthijs: Needs alphabetic ordering

node_modules/oae-core/filepreview/css/filepreview.css
```diff
@@ -90,4 +90,11 @@
...
+    left: 20px !important;
```
nicolaasmatthijs: It's better if we can avoid the `!important` statements. A better approach would be to put a `#filepreview-container` quantifier in front of out so the rules are limited to this widget and overrides happen correctly.

node_modules/oae-core/filepreview/css/filepreview.css
```diff
@@ -90,4 +90,11 @@
...
+}
+
+.video-js .vjs-control-bar {
```
nicolaasmatthijs: This doesn't seem like the best approach to make the widgets align. The best approach would be to add padding on the left and right of the video player to make it align with the comments widget.

nicolaasmatthijs:
Added some more review. Back to @sallakarppinen

node_modules/oae-core/filepreview/css/filepreview.css
```diff
@@ -83,23 +83,25 @@
     }
 }
 
+/* Video.js overrides. These are required to align the video
+ element with the comment widget and to get it to resize correctly */
+
 .video-js {
     position: unset;
     background-color: #fff;
```
nicolaasmatthijs: Needs to be capitalized

node_modules/oae-core/filepreview/css/filepreview.css
```diff
@@ -83,23 +83,25 @@
...
 }
 
-.vjs-tech {
-    background-color: #000;
-    width: 98% !important;
-    left: 20px !important;
+#filepreview-video-container .video-js .vjs-control-bar {
+    width: 98%;
```
nicolaasmatthijs: I'm nervous about the 98% width as it feels like it might not always align well with the comments widget. My preference would be to find an approach where we just add the appropriate margin to the video player container and then let it take up the space it has available.

node_modules/oae-core/filepreview/css/filepreview.css
```diff
@@ -83,23 +83,25 @@
...
 }
 
-.vjs-poster {
-    left: 20px !important;
-    right: 20px !important;
-}
\ No newline at end of file
+#filepreview-video-container .vjs-tech {
+    background-color: #000;
+    width: 98%;
```
nicolaasmatthijs: Properties need to be sorted alphabetically

node_modules/oae-core/filepreview/css/filepreview.css
```diff
@@ -83,23 +83,25 @@
...
     background-color: #fff;
 }
 
-.vjs-tech {
-    background-color: #000;
-    width: 98% !important;
-    left: 20px !important;
+#filepreview-video-container .video-js .vjs-control-bar {
+    width: 98%;
+    margin: 0 20px;
```
nicolaasmatthijs: Alphabetic sorting

nicolaasmatthijs:
Finished another round of review. Back to @sallakarppinen

node_modules/oae-core/filepreview/css/filepreview.css
```diff
@@ -58,39 +58,50 @@
...
+
+/* Video.js overrides. These are required to align the video
+ element with the comment widget and to get it to resize correctly */
+
+.video-js {
+    background-color: #FFF;
+    position: unset;
+}
+
+#filepreview-video-container .video-js .vjs-control-bar {
```
nicolaasmatthijs: We can write this a bit more efficiently:

```
#filepreview-video-container .video-js .vjs-control-bar,
#filepreview-video-container .vjs-tech {
  margin: 0 20px;
  position: unset;
}

#filepreview-video-container .vjs-poster {
    margin: 0 20px;
}

#filepreview-video-container .vjs-tech {
    background-color: #000;
 }
```

node_modules/oae-core/filepreview/css/filepreview.css
```diff
@@ -58,39 +58,50 @@
...
+    margin: 0 20px;
```
nicolaasmatthijs: Note that having to set this `margin` on the different components is usually an indication that the `margin` can be set once on the container of these components. Is that possible here?

nicolaasmatthijs:
Left some final comments. Back to @sallakarppinen

sallakarppinen:
Made some changes (including tidying up repetition as brought up by @nicolaasmatthijs)... Had some weird behaviour on qa2, it seems that using `position: unset` caused the video to not play properly so I've swapped back to `width: 98%`.

mrvisser:
Looks like the "play" icon for MP3 is a bit out of wack:

image

mrvisser:
There are also issues when you play it with the player UI, and the file only seems to play the first half a second or so of the file then stops:

image

nicolaasmatthijs:
The following functional issues were identified during the bugbash:

1) The `Processing this file` screen is not correctly updated once the video has finished transcoding and requires a refresh of the screen before the video can be seen. We expect a push notification to inform the UI that it has finished processing and update the content preview. This is in place for other types of content already.

screen shot 2016-08-25 at 15 17 19

2) The video player needs to align on both ends with the comments widget at all widths. That's why I was pushing back on the `98%` width as I can't see how that can ensure the alignment. In the screenshot below, you'll see red lines where the video is expected to align:

screen-shot-2016-08-25-at-15 17 39

sallakarppinen:
@nicolaasmatthijs: For issue 2, could you specify the browser and OS that were used, since I'm  not able to reproduce (on Firefox/Linux Mint)?

nicolaasmatthijs:
@sallakarppinen : It appears to be any browser (Firefox / Chrome / Safari). Maybe try changing the width of your browser window as there's a single "perfect" width where they do roughly align (because of the percentage width).

sallakarppinen:
@nicolaasmatthijs You're right, I can see it if I resize to something other than the 'standard' sizes... I'll try to find a way around it.

sallakarppinen:
Have not been able to fix the resizing, but the first issue (file not refreshing when processed) is fixed by the latest changes to Hilary branch for this task which is deployed on qa2.

PR #4132, opened 2017-04-05
Clean out IE 9-specific CSS hacks

fortMaximus:


fortMaximus:
@sallakarppinen So should I submit another PR (and add the edited JS/HTML files to it)?

sallakarppinen:
You can update this PR by making more changes on your branch and pushing them. :)

node_modules/oae-core/upload/js/upload.js
```diff
@@ -35,9 +35,6 @@ define(['jquery', 'oae.core', 'jquery.fileupload', 'jquery.iframe-transport', 'j
         // will be used in the event communication between this widget and the `setpermissions` widget.
         var setPermissionsId = oae.api.util.generateId();
 
-        // IE9 and below don't support XHR file uploads and we fall back to iframe transport
-        var useIframeTransport = !$.support.xhrFileUpload && !$.support.xhrFormDataFileUpload;
```
sallakarppinen: We still reference this variable elsewhere in this file - those references should be updated too.

node_modules/oae-core/upload/js/upload.js
```diff
@@ -478,9 +467,6 @@ define(['jquery', 'oae.core', 'jquery.fileupload', 'jquery.iframe-transport', 'j
                 }
 
                 // Disable editing on upload
-                // Note: the file input element can not be disabled, as that will cause IE9
-                // to drop it from the  DOM and not submit its file content to the server
-                $('#upload-modal *').not('input[type="file"]').prop('disabled', true);
```
sallakarppinen: This shouldn't be taken out... What it's doing is disabling the upload modal when we're already in progress. We could change it to `$('#upload-modal *').prop('disabled', true);` though since we're no longer limited by IE9 constriction on `input[type="file"]`.

shared/oae/api/oae.api.util.js
```diff
@@ -1866,7 +1866,5 @@ define(['exports', 'require', 'jquery', 'underscore', 'oae.api.config', 'markdow
      *
      * @return {String}   Event(s) that can fire when the content of an input field changes
      */
-    var getFormChangeEventNames = exports.getFormChangeEventNames = function() {
-        return $('html').hasClass('ie-lt10') ? 'change keyup paste cut' : 'change input';
-    };
+    var getFormChangeEventNames = exports.getFormChangeEventNames = return "change input";
```
sallakarppinen: This is not valid syntax... An export needs to be a function, so this would have to be `var getFormChangeEventNames = exports.getFormChangeEventNames = function() {return "change input";}`. However, in this case we shouldn't need the function at all anymore since it's a IE9 workaround, so a better solution would be to take out the function completely and change all references to it with the string `'change input'`, like mentioned in the TODO comment.

fortMaximus: @sallakarppinen It seems that I missed up the JS files. So should I leave the JS related IE9 hacks to someone else and return to the previous PR?

sallakarppinen: I don't think you messed it up, it just needs a bit of improvement... But you can leave the JS to me as well if you'd rather do that. :smile:

fortMaximus: I just reread your comments and realized they're easy to fix, so I will do it.

sallakarppinen: Glad to hear that, and please ask if you have any questions!

node_modules/oae-core/changepic/js/changepic.js
```diff
@@ -36,9 +36,6 @@ define(['jquery', 'oae.core', 'jquery.jcrop', 'jquery.fileupload', 'jquery.ifram
             principalId: null
         };
 
-        // IE9 and below don't support XHR file uploads and we fall back to iframe transport
-        var useIframeTransport = !$.support.xhrFileUpload && !$.support.xhrFormDataFileUpload;
```
fortMaximus: When there's some code like this one:
```
if (!useIframeTransport) {
                        // Update the progress bar
                        updateProgress((data.loaded / data.total) * 100);
                    }
```
Should I remove the `if` statement line only and keep the code inside its blocks?

shared/oae/api/oae.api.util.js
fortMaximus: About this file: I didn't find any references for the function `getFormChangeEventNames` to replace.

node_modules/oae-core/changepic/js/changepic.js
sallakarppinen: Yes, exactly, so for this one it'd be
```javascript
'progress': function(ev, data) {
  updateProgress((data.loaded / data.total) * 100);
}
```

shared/oae/api/oae.api.util.js
sallakarppinen: The references will be in other files where this one is `require`d from. You can find them by using your IDEs search function or by `grep`ping the code base.

fortMaximus: I didn't know that there was a universal search function in my editor (that will save me a lot of time). The way I did it before was to search every file in project for a keyword.

Also, just read about `grep`.

Thanks for the tip.

node_modules/oae-core/lhnavigation/js/lhnavigation.js
```diff
@@ -171,15 +171,6 @@ define(['jquery', 'oae.core', 'jquery.history'], function($, oae) {
                 '_': Math.random()
             };
 
-            // Although the `title` and `url` field are optional, we need to enter them as IE9 simply doesn't
-            // handle their absense very well. We cannot lose any parameters such as the `q` query string
-            // parameter as IE9 is not able to recover it from its state.
-            var title = selectedPage.title;
-            if (query) {
```
fortMaximus: So I need to only delete the code between the `if` statement brackets here?

sallakarppinen: Hmm, no. I don't actually think any of the code here needs to be deleted... We should keep `query` in `data.query`, so we should keep that part from inside the brackets, and there's no harm in keeping the title or adding the query to `loadedPath` either.

shared/oae/api/oae.api.util.js
fortMaximus: Hey, I don't know what I've done.
I was working on the master branch (instead of the branch I created) and I committed and pushed my code there also, then I tried to delete it (using `git reset --hard <sha1-commit-id>` then `git push origin HEAD --force`) but I don't think that's it's deleted (because it still says (in the master branch) this branch is 1 commit ahead and my commit shows in the "latest commit" also).

sallakarppinen: Hmm... I think that should've worked, but I can see the extra commit on your master as well so apparently it didn't... What output did you get from the `git push origin HEAD --force` command? If your local master branch is correct now, try `git push origin master --force`?

brecke: Hi,

I'm not sure how I can be of help without some extra information. If you do `git reset` that will change the HEAD pointer to a different commit, and if you run `git push origin master` afterward, you'll get the same effect in the remote repo. Was this what you did? These steps delete nothing, every commit is still there.

But perhaps it would help if you told us what it was you tried to achieve in the first place?

davidoae: Might be easier to re-clone.

fortMaximus: @sallakarppinen 
> What output did you get from the `git push origin HEAD --force` command?

Here's the output after executing that command:
```
Counting objects: 36, done.
Delta compression using up to 8 threads.
Compressing objects: 100% (27/27), done.
Writing objects: 100% (36/36), 2.81 KiB | 0 bytes/s, done.
Total 36 (delta 24), reused 0 (delta 0)
remote: Resolving deltas: 100% (24/24), completed with 24 local objects.
To https://github.com/fortMaximus/3akai-ux
 23a0d2d..68317d6  HEAD -> master
```
>  If your local master branch is correct now, try `git push origin master --force`?

Can you explain why I should do this?

sallakarppinen: @brecke, that output is different from what you got when trying this locally, correct? Different configs perhaps?

> Can you explain why I should do this?

Doing that means you would push your local master branch to the remote master branch (using force as the remote branch is now technically 'ahead), which I think is what you're trying to do?

fortMaximus: @brecke 
>Was this what you did? 

These are the exact commands I used `git reset --hard <sha1-commit-id>` then `git push origin HEAD --force`

>But perhaps it would help if you told us what it was you tried to achieve in the first place?

I thought that I was working on the branch I created, but I discovered after committing and pushing my changes that I was working on the master. So, I wanted to reverse what I did.

brecke: Thanks for that. Well, I think you managed to achieve what you intended, in the master branch. But your real problem is how to apply those extra commits to the feature branch you're working on, is this correct? Currently it should be similar to this:

screen shot 2017-04-07 at 14 33 58

If so, you can now use cherry picking, which would go like:

screen shot 2017-04-07 at 14 34 41

...or you can just `git reset` back to your latest commit on `master` and then merge that branch onto the feature branch. It would go something like this:

screen shot 2017-04-07 at 14 42 45

Hope that helps!

fortMaximus: @brecke 
Thank you.

>Well, I think you managed to achieve what you intended, in the master branch.

No, the problem is still in the master branch, it still has the commit (that I wanted to push to the feature branch) but pushed it to the master instead.

sallakarppinen: Did you try my suggestion above? 😒

brecke: @fortMaximus well then, I guess there are two steps now:

1. Make sure you cherry-pick that commit to the feature branch (see second image) and then push it. This should be enough for you to keep working.
2. Delete that commit from master branch. What you did should have worked, but there's probably something I'm not grasping. Can you see that commit in `git log --oneline`? or does it appear in `git reflog`?

The first image above should picture your current status, doesn't it?

fortMaximus: @brecke 
> From the masterbranch, can you see that commit in git log --oneline?

Yes.

>The first image above should picture your current status, doesn't it?

I am not sure, because I still don't understand what's exactly going on.

fortMaximus: I think I'll leave fixing the IE9 specific JS tricks for later or for someone else to take it (after I learn more about Git).

sallakarppinen: That's fine... In that case, do you think you could reset the `clean-out-ie9-css-hacks` branch to the CSS commit and re-push it? You should be able to do that with `git reset --hard d34cabe130ee0a64141030bec44cad2600ef45e6` and then `git push origin clean-out-ie9-css-hacks --force`?

fortMaximus: @sallakarppinen done.

sallakarppinen:
The JS commit was still showing up on this branch so I removed it - good job, @fortMaximus, merged. :smile:

fortMaximus:
@sallakarppinen @brecke Thank you for the support. I'm happy, this my first pull request and it got merged too. :tada:

PR #4133, opened 2017-04-05
Docker-based development environment

brecke:
Frontend follow-up for PR [#1349](https://github.com/oaeproject/Hilary/pull/1349)

### DIFF
--- a/nginx/nginx.conf.docker
+++ b/nginx/nginx.conf.docker
@@ -194,6 +194,7 @@ http {
 
         location /api/ui/staticbatch {
             expires max;
+            proxy_read_timeout 300;
             proxy_pass http://globaladminworkers;
         }
 
@@ -529,6 +530,7 @@ http {
 
         location /api/ui/staticbatch {
             expires max;
+            proxy_read_timeout 300;
             proxy_pass http://tenantworkers;
         }
 
@@ -588,6 +590,7 @@ http {
         # Explicitly don't cache any other API requests
         location /api/ {
             expires -1;
+            proxy_read_timeout 300;
             proxy_pass http://tenantworkers;
         }
 
--- a/nginx/nginx.conf.docker
+++ b/nginx/nginx.conf.docker
@@ -358,8 +358,6 @@ http {
         rewrite ^/search        /ui/search.html last;
         rewrite ^/signup        /ui/signup.html last;
         rewrite ^/user          /ui/user.html last;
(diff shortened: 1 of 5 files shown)


nginx/nginx.conf.docker
```diff
@@ -0,0 +1,637 @@
...
+        rewrite ^/discussion    /ui/discussion.html last;
+        rewrite ^/folder        /ui/folder.html last;
+        rewrite ^/groups        /ui/index.html last;
+        rewrite ^/group         /ui/group.html last;
+        rewrite ^/library       /ui/index.html last;
+        rewrite ^/network       /ui/index.html last;
+        rewrite ^/search        /ui/search.html last;
+        rewrite ^/signup        /ui/signup.html last;
+        rewrite ^/user          /ui/user.html last;
+        rewrite ^/resetpassword$    /ui/resetpassword.html last;
```
sallakarppinen: This PR has not been merged yet.

brecke: oops.

brecke:
> firstly, do we need to keep both nginx.conf files (including making sure they're updated as needed etc) in sync for the foreseeable future? I assume this will be the case since the other one is required for the manual setup...

Exactly.

> Secondly, can we support local https with the docker setup as will be required by, for example, the jitsi PR in the future?

Yes, pretty sure it's possible. Should we do it now?

sallakarppinen:
> Yes, pretty sure it's possible. Should we do it now?

Well, either we do it now or we do it in a (...hopefully) short while when the Jitsi stuff is ready to be merged. :smile:

brecke:
☝️ HTTPS support via dcd3a69 (optional)

nginx/ssl-params.conf
```diff
@@ -0,0 +1,21 @@
...
+ssl_session_tickets off;
+ssl_stapling on;
+ssl_stapling_verify on;
+resolver 8.8.8.8 8.8.4.4 valid=300s;
+resolver_timeout 5s;
+# Disable preloading HSTS for now.  You can use the commented out header line that includes
+# the "preload" directive if you understand the implications.
+#add_header Strict-Transport-Security "max-age=63072000; includeSubdomains; preload";
+add_header Strict-Transport-Security "max-age=63072000; includeSubdomains";
+add_header X-Frame-Options DENY;
```
sallakarppinen: This might break etherpad?

brecke: You are right, it breaks some stuff. I still needs some work.

PR #4150, opened 2017-05-12
Front-end for LTI tool integration

sallakarppinen:
WIP - still missing launching and displaying tools.


node_modules/oae-core/lti/deletelti/js/deletelti.js
```diff
@@ -0,0 +1,276 @@
...
+
+            $rootel.on('click', '#deletelti-manage-delete-system', function() {
+                deleteLtiTools();
+            });
+        };
+
+        setUpReset();
+        setUpDeleteLtiModal();
+    };
+});
```
brecke: +line

shared/oae/api/oae.api.lti.js
```diff
@@ -0,0 +1,150 @@
...
+            'type': 'DELETE',
+            'success': function(data) {
+                callback();
+            },
+            'error': function(jqXHR, textStatus) {
+                callback({'code': jqXHR.status, 'msg': jqXHR.responseText});
+            }
+        });
+    };
+});
```
brecke: +line

shared/oae/css/oae.components.css
```diff
@@ -503,11 +504,6 @@ div.oae-thumbnail.fa-oae-collabdoc:before {
     content: "\f044";
 }
 
-div.oae-thumbnail.fa-oae-folder:before {
```
brecke: why have this removed?

ui/js/lti.js
```diff
@@ -0,0 +1,90 @@
...
+                    oae.api.util.redirect().notfound();
+                }
+                return;
+            }
+            toolProfile = data;
+            var launchParams = toolProfile.launchParams;
+            var tool = toolProfile.tool;
+
+            oae.api.util.setBrowserTitle(tool.displayName);
+            var form = '<form method="POST" id="lti-form" action="' + tool.launchUrl + '" target="lti-name">';
```
brecke: I know this works, but why not use jQuery for this? It would go something like this:

```
var form = $("<form/>", { action:'/myaction' });

form.append(
    $("<input>", 
         { type:'text', 
           placeholder:'Keywords', 
           name:'keyword', 
           style:'width:65%' }
     )
);

form.append( 
     $("<input>", 
          { type:'submit', 
            value:'Search', 
            style:'width:30%' }
       )
);

$("#someDivId").append(form);
```

(taken from http://stackoverflow.com/questions/4497140/jquery-create-a-form-and-add-elements-to-it-programmatically)

This kind of solution looks way better than dealing with strings.

shared/oae/css/oae.components.css
sallakarppinen: It was in twice; repeats line 499.

PR #4157, opened 2017-05-30
Fixed XSS vulnerability for front-end

sallakarppinen:


sallakarppinen:
I'll merge it and we can separate them to their own function in the future if we'll need to.

PR #4172, opened 2017-08-31
Added another dependency to exclusion regex

sallakarppinen:


sallakarppinen:
I've looked into replacing current require.js optimisation with uglify-es, which is the ES6-aware version of uglify as well... It would probably make sense to do that as a part of a larger update to 3akai grunt settings, where we also go over the current dependencies and update/remove them as appropriate. I don't think this should be a part of the next release but I'll create an issue in the backlog for it for when we have 'extra time'.

Issue #4177, opened 2017-10-02
Ability to share links among tenancies


brecke:
Sister issue to [1449](https://github.com/oaeproject/Hilary/issues/1449).

Tenancy pickers available:

- http://unity.ac/
- http://oaeproject.org/

brecke:
PR https://github.com/oaeproject/3akai-ux/pull/4185

PR #4185, opened 2017-12-19
Issue 4177

brecke:
Corresponds to https://github.com/oaeproject/3akai-ux/issues/4177


node_modules/oae-core/signin/signin.html
```diff
@@ -1,81 +1,126 @@
 <!-- CSS -->
 <link rel="stylesheet" type="text/css" href="css/signin.css" />
+<link rel="stylesheet" type="text/css" href="css/select2.css" />
```
sallakarppinen: As mentioned previously, this should really be in the shared folder - otherwise we won't be able to use the lib elsewhere.

node_modules/oae-core/signin/signin.html
```diff
@@ -1,81 +1,126 @@
...
-                </div>
-                <div class="form-group">
-                    <button type="submit" class="btn btn-primary">__MSG__SIGN_IN__</button>
-                </div>
-            </form>
+    </div>
+	<h4 id="signin-modal-title">__MSG__SIGN_IN_TO_OTHER_TENANCY__</h4>
+    <div id="tenancyForm">
+        <select tabindex="2" id="tenancyInput">
+            <option value="3620194" selected="selected"> - Type your institution -</option>
```
sallakarppinen: Is there a reason for this specific option value? I'd prefer using something like `---` instead for the empty default.

shared/oae/api/oae.api.authentication.js
```diff
@@ -319,6 +319,26 @@ define(['exports', 'jquery', 'oae.api.config', 'oae.api.i18n', 'oae.api.user', '
         });
     };
 
+    /**
+     * TODO bla bla bla
```
sallakarppinen: ...

shared/oae/api/oae.api.authentication.js
```diff
@@ -319,6 +319,26 @@ define(['exports', 'jquery', 'oae.api.config', 'oae.api.i18n', 'oae.api.user', '
...
+     */
+    var removeLoggedInTenanciesCookie = exports.removeLoggedInTenanciesCookie = function(currentTenant, callback) {
+        // let's add this tenant to the loggedin_tenancies cookie
```
sallakarppinen: Remove, not add?

shared/oae/api/oae.api.authentication.js
```diff
@@ -319,6 +319,26 @@ define(['exports', 'jquery', 'oae.api.config', 'oae.api.i18n', 'oae.api.user', '
...
+        let cookieName = 'loggedin_tenancies';
+        let loggedInTenancies = docCookies.getItem(cookieName);
+        loggedInTenancies = loggedInTenancies ? JSON.parse(loggedInTenancies) : [];
+        loggedInTenancies = _.reject(loggedInTenancies, (eachTenancy) => {
+            return eachTenancy.alias === currentTenant.alias;
+        });
+
+        // set cookie with updated data
+        let location = $.url(encodeURI(window.location.href));
+		let host = location.attr('host');
```
sallakarppinen: Slightly weird whitespace...

shared/oae/api/oae.api.tenancysearch.js
```diff
@@ -0,0 +1,139 @@
...
+                                return {
+                                    host: eachResult.host,
+                                    displayName: eachResult.displayName,
+                                    id: eachResult.alias,
+                                    code: eachResult.countryCode
+                                };
+                            });
+
+                            tenancyResults = tenancyResults.slice(0, 10);
+                            tenancyResults.sort(function (a, b) {
```
sallakarppinen: We sort results by length of display name? 😕

shared/oae/api/oae.api.tenancysearch.js
```diff
@@ -0,0 +1,139 @@
...
+                this.$alert = this.$module.find('#tenancyAlert');
+            },
+            bindEvents: function() {
+                this.$input.on('select2:select', this.tenancySelect.bind(this));
+            },
+            render: function() {
+                this.$input.select2({
+                    dropdownParent: $('#signin-modal'),
+                    ajax: {
+                        url: '/api/search/tenants?limit=2000',
```
sallakarppinen: Will 2000 always be enough?

shared/oae/api/oae.api.util.js
```diff
@@ -1804,14 +1804,25 @@ define(['exports', 'require', 'jquery', 'underscore', 'oae.api.config', 'markdow
             window.location = '/servermaintenance';
         };
 
+        /**
+         * TODO bla bla bla
```
sallakarppinen: ...

shared/oae/api/oae.api.util.js
```diff
@@ -1804,14 +1804,25 @@ define(['exports', 'require', 'jquery', 'underscore', 'oae.api.config', 'markdow
...
+         */
+        var logInThroughAnotherTenant = function(tenant) {
+            let relative = url().attr('query').slice(4);
+            let host = url().attr('base').replace(/(\w*)\./, tenant.alias + '.');
+            let newURL = host + relative;
+            window.location = newURL;
```
sallakarppinen: Will this continue to work when the tenants are on different domains (for example, *.unity.ac to *.oaeproject.org)?

shared/oae/api/oae.core.js
```diff
@@ -84,7 +85,9 @@ define([
         'markdown',
         'sockjs',
         'tinycon',
-        'underscore'
+        'underscore',
+        'moment',
+        'cookies'
```
sallakarppinen: Do we also need select2&async here?

node_modules/oae-core/signin/js/signin.js
```diff
@@ -133,24 +168,104 @@ define(['jquery','oae.core'], function($, oae) {
             });
         };
 
+        /**
+         * TODO bla bla bla
```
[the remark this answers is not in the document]
brecke: oops

node_modules/oae-core/signin/js/signin.js
```diff
@@ -133,24 +168,104 @@ define(['jquery','oae.core'], function($, oae) {
...
+         */
+        let _getTenantLogo = function (tenantAlias, callback) {
+            $.ajax({
+                'url': `${protocol}://${tenantAlias}.oae.com/api/ui/logo`,
```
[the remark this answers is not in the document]
brecke: Thank you for spotting that 🤦‍♂️ 

The problem with relative URL in this specific scenario is that the `/api/ui/logo` endpoint returns the logo for the tenant in `tenantAlias.domain/api/ui/logo`, hence doing it this way. It does sound a little tricky, but it's the easiest way IMO.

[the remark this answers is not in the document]
sallakarppinen: That makes sense... `tenant.host` might have the right URL?

node_modules/oae-core/signin/js/signin.js
```diff
@@ -133,24 +168,104 @@ define(['jquery','oae.core'], function($, oae) {
...
+
+            /**
+             * Steps
+             * 1 Fetch which other tenants the user is logged in to through a cookie
+             * 2 Fetch other tenants logos from the server
+             * 3 Present a modal box with options to fetch the same URL through a different tenant
+             */
+
+            let allCookies = document.cookie;
+            let currentTenant = oae.data.me.tenant;
```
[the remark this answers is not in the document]
brecke: That variable returns the tenant the user is visiting, not the one he's currently logged in as far as I understand it.

screen shot 2017-12-20 at 11 26 15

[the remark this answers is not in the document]
sallakarppinen: Oh, so it does. Right, sorry.

node_modules/oae-core/signin/js/signin.js
```diff
@@ -133,24 +168,104 @@ define(['jquery','oae.core'], function($, oae) {
...
+            let loggedInTenancies = docCookies.getItem(cookieName);
+            loggedInTenancies = loggedInTenancies ? JSON.parse(loggedInTenancies) : [];
+
+            // Remove own tenant from tenant list
+            loggedInTenancies = _.reject(loggedInTenancies, (eachTenancy) => {
+                return eachTenancy.alias === currentTenant.alias;
+            });
+
+            // sort them by the last login time
+            loggedInTenancies = _.sortBy(loggedInTenancies, 'lastLogin').reverse();
```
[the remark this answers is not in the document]
brecke: Well spotted. I think this made sense before I removed the cookie on logout... 👍

node_modules/oae-core/signin/js/signin.js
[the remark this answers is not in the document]
brecke: there's no the final version is 

```
const host = location.attr('host');
const dotDomain = '.'.concat(host.split('.').slice(1).join('.'));

...

'url': `${protocol}://${tenantAlias}${dotDomain}/api/ui/logo`,
```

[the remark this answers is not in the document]
sallakarppinen: 👍

node_modules/oae-core/signin/signin.html
brecke: Copy paste ✂️

shared/oae/api/oae.api.tenancysearch.js
brecke: how likely is it going to be having the search query resulting in over 2000 entries? 🤷‍♂️

sallakarppinen: Pretty unlikely, I admit.

shared/oae/api/oae.api.tenancysearch.js
brecke: It makes no sense, I know, but that's how the unity.ac works, and I just thought there must have been a reason so I left it the same. Any ideas on how to do this better? maybe sort alphabetically?

shared/oae/api/oae.api.util.js
brecke: hmm no, I don't think so. Which means I might need to do some changes such as storing not just the tenantAlias but also the tenant domain 

🖐

shared/oae/api/oae.core.js
brecke: I honestly don't know. It works without it, loading those selectively when needed 🤷‍♂️

shared/oae/api/oae.api.tenancysearch.js
sallakarppinen: I think alphabetical would be better... I don't know what the reasoning there was but to me it doesn't really make sense.

shared/oae/api/oae.api.tenancysearch.js
brecke: well, the user will have to type in more than that. It's like googling "university" trying to find a specific one, except that this is just a dropdown picker thus cannot have proper pagination, only limited scroll... I think that the 2000 limit is more of a protection for the DB query more than anything else. Any ideas? 🤷‍♂️

sallakarppinen: Yeah, obviously it'd be a bad query, but I can still totally see someone doing it... 🤷‍♀️ I also noticed that unity.ac is actually pretty bad with queries like that (try typing in 'university of' and then pausing for example). That's probably more a sorting problem though.

brecke: plus we have this: `tenancyResults = tenancyResults.slice(0, 10);` ... which makes the 2000 kind of obsolete.

I think that it's implicit that the user needs to search something quite specific in order to find the intended tenancy. The display (dropdown picker) doesn't scale to a lot of results, but listing +2000 results isn't something we'd want to do either way - not even paged - cuz it would no longer be an improvement over what we have at the moment.

brecke: It is, yes :)

shared/oae/api/oae.api.util.js
brecke: OK, so I did a bit of research on this and the reason why the scenario won't work is because of the way cookies work. If user logs in to `tenantA.unity.ac`, the cookie domains one can set are either `.tenantA.unity.ac` and `.unity.ac`. The latter is the one that's useful to us, so that other tenants can access it (e.g. `tenantB.unity.ac`). But one could never set a cookie for `tenantC.oaeproject.org` to see, because the domains are different. (Read [this thread](https://stackoverflow.com/questions/1062963/how-do-browser-cookie-domains-work) for more info.)

So that means this is indeed a limitation of this approach, and I'm guessing that all other kinds of browser storage respect the same philosophy for security reasons.

sallakarppinen: Fair enough, I hadn't thought of that. That's an unavoidable limitation then, but I guess generally most tenancies in any instance would be using the same main domain.

sallakarppinen:
Hmm. On Chrome, I seem to get stuck on the redirect? (Steps: pasted URL of a private content item in a private window, searched for my tenancy in the login prompt, clicked on the tenancy.)

brecke:
Cannot reproduce :-/ What am I missing?

<a href="https://cl.ly/0V2P0B2C102W" target="_blank"></a>

sallakarppinen:
Turns out it was a problem with my local setup, so I deleted the comment... Apparently not fast enough though. 😬 Sorry!

I've now tested the happy path (user has access on second tenancy) and sad path (user does not have access on second tenancy) plus searching for tenancies... All seems to work. Am doing a final comb through the code next.

node_modules/oae-core/signin/js/signin.js
```diff
@@ -53,6 +59,33 @@ define(['jquery','oae.core'], function($, oae) {
...
+                loggedInTenancies = loggedInTenancies ? JSON.parse(loggedInTenancies) : [];
+                let currentTenant = oae.data.me.tenant;
+                let existingCookie = _.findWhere(loggedInTenancies, { alias: currentTenant.alias });
+
+                // a new cookie to be stored
+                let newCookie = (function() {
+                    return {
+                        'alias': currentTenant.alias,
+                        'displayName': currentTenant.displayName,
+                        'lastLogin': Date.now()
```
sallakarppinen: Still using `Date.now()` here and `moment()` elsewhere - either one is fine, but we shouldn't mix the two.

node_modules/oae-core/signin/js/signin.js
```diff
@@ -133,24 +168,104 @@ define(['jquery','oae.core'], function($, oae) {
...
+            /**
+             * Steps
+             * 1 Fetch which other tenants the user is logged in to through a cookie
+             * 2 Fetch other tenants logos from the server
+             * 3 Present a modal box with options to fetch the same URL through a different tenant
+             */
+
+            let allCookies = document.cookie;
+            let currentTenant = oae.data.me.tenant;
+            let now = moment();
```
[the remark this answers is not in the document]
sallakarppinen: ☝️

node_modules/oae-core/signin/js/signin.js
```diff
@@ -133,30 +166,107 @@ define(['jquery','oae.core'], function($, oae) {
...
+                // Handle username / password authentication
+                initLocalSignInFormValidation();
+            });
         };
 
         /**
          * Render the footer for the signin modal
          */
-        var renderSignInFooter = function() {
+        let renderSignInFooter = function() {
```
sallakarppinen: Could be `const`?

brecke:
👍 bash at will!

shared/oae/api/oae.api.authentication.js
```diff
@@ -319,10 +319,33 @@ define(['exports', 'jquery', 'oae.api.config', 'oae.api.i18n', 'oae.api.user', '
...
+    /**
+     * Remove a specific tenant from the cookie where we keep track of tenants the user logged in to
+     *
+     * @param  {String}   currentTenant   The current tenant the user is visiting
+     * @param  {Function} callback        Standard callback function
+     */
+    const removeLoggedInTenanciesCookie = exports.removeLoggedInTenanciesCookie = function(currentTenant, callback) {
+        // let's remove this tenant to the loggedin_tenancies cookie
+        let cookieName = 'loggedin_tenancies';
+        let loggedInTenancies = docCookies.getItem(cookieName);
```
sallakarppinen: Where is `docCookies` coming from? I don't seem to find where we fetch the cookies.

shared/oae/api/oae.api.tenancysearch.js
```diff
@@ -0,0 +1,129 @@
...
+                                };
+                            });
+
+                            tenancyResults = _.sortBy(tenancyResults, 'displayName');
+                            params.page = params.page || 1;
+
+                            return {
+                                results: tenancyResults,
+                                pagination: {
+                                    more: (params.page * 30) < data.totalCount
```
sallakarppinen: `* 30`?

shared/oae/api/oae.core.js
```diff
@@ -84,7 +85,8 @@ define([
         'markdown',
         'sockjs',
         'tinycon',
-        'underscore'
+        'underscore',
+        'cookies'
```
sallakarppinen: I'd still be happier if we had the rest of the external libs here too... (Also, I think these should be alphabetical.)

node_modules/oae-core/signin/css/signin.css
```diff
@@ -74,3 +74,97 @@ body.oae-page-signup #signin-modal-signup, body.oae-page-signup #signin-modal-re
 #signin-modal-resetpassword > button {
     float: right;
 }
+
+select#form_field, .select2-container--default {
+	width: 100% !important;
```
sallakarppinen: Are all the `!important`s necessary?

brecke: Apparently, yes. There's a comment somewhere stating explicitly that `!important` should be kept, but I'm not an expert in CSS to assess if that's really the case. Is it a bad practice or something?

node_modules/oae-core/signin/js/signin.js
brecke: Don't understand why we can't have both, since they serve different purposes. What am I missing here?

node_modules/oae-core/signin/js/signin.js
brecke: 👍

shared/oae/api/oae.api.authentication.js
brecke: it's a small lib I've included in `/shared/vendor/js` and is being imported in `oae.core.js`

sallakarppinen: Isn't it just `cookies` in oae.core.js?

node_modules/oae-core/signin/js/signin.js
sallakarppinen: In this case they seem to be serving the same purpose though - logging the current time? It's mainly this line that I have problems with `eachTenancy.lastLoginFromNow = moment(parseInt(eachTenancy.lastLogin)).from(now);` - here the `eachTenancy.lastLogin` was recorded as `Date.now()` while `now` was using `moment()` - while it works fine, it just seems messy to me.

shared/oae/api/oae.api.tenancysearch.js
brecke: oops. it's `10` now, to match the number of results being shown.

shared/oae/api/oae.core.js
brecke: Alphabetical order 🆗 

What other external libs? Are we missing some?

node_modules/oae-core/signin/css/signin.css
sallakarppinen: I think it's generally considered pretty poor practice since it kind of overrides the 'cascading' part of CSS. Can't be helped though I guess?

ui/js/folder.js
```diff
@@ -13,7 +13,7 @@
  * permissions and limitations under the License.
  */
 
-require(['jquery', 'underscore', 'oae.core'], function($, _, oae) {
+require(['jquery', 'underscore', 'oae.core', 'moment', 'async'], function($, _, oae, moment, async) {
```
sallakarppinen: Do we actually use moment/async here?

shared/oae/api/oae.core.js
sallakarppinen: > What other external libs? Are we missing some?

Ignore me, I thought we were in the bootstrap file. 🤦‍♀️

shared/oae/api/oae.api.authentication.js
brecke: It is, but it exports a variable `docCookies`

sallakarppinen: Righto.

node_modules/oae-core/signin/js/signin.js
brecke: I can have `now = new Date()` instead and use it in the `moment#Constructor` but that's probably slower (compared to `moment()`) and more verbose. I'll test that.

ui/js/folder.js
brecke: hehe I used to, but then refactored and moved code away. It's been removed now.

node_modules/oae-core/signin/js/signin.js
sallakarppinen: Couldn't we just use `moment()` here?

brecke: Well, `moment()` returns a `Moment` object which may be used to parse or manipulate dates but all I need is a serializable Date to store the cookie. Why would I create an object when a number suffices?

```
-> Date.now();
1513871496419 # this is what I need
-> moment();
moment("2017-12-21T15:51:40.229")
```

sallakarppinen: Do we really need  a moment object later when getting the difference between `lastLogin` and `now` later in the code, then?

brecke: According to the docs, the `moment#from` method accepts both a `Moment` object or a `Date` object. I had picked the less verbose and more natural way - using the former - because it seemed simpler not mixing objects (`moment.from(moment)` as opposed to `moment.from(date)`). But it's changed now.

sallakarppinen: 👍

PR #4191, opened 2018-01-22
Added sorting and option to search items by me

sallakarppinen:
Closes #4180

PR #4193, opened 2018-01-22
fixed circleci script, hopefully

### LINKED ISSUE #4192: CircleCI tests aren't passing

brecke:
Fixes #4192

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,13 +1,30 @@
 version: 2
 jobs:
   build:
-    machine: true
+    docker:
+      - image: mhart/alpine-node:8.9.4
     working_directory: ~/3akai-ux
     steps:
+      - run:
+          name: Install dependencies
+          command: >
+            apk add --update --no-cache
+            git
+            openssh-client
       - checkout
-      - run: npm install -g npm@3.10.10 # default npm version is 5 and it doesn't work
-      # install dependencies
-      - run: npm install
-      # check grunt build
-      - run: grunt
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -1,13 +1,28 @@
 version: 2
 jobs:
   build:
-    machine: true
+    docker:
+      - image: mhart/alpine-node:8.9.4
```
sallakarppinen: This is a different alpine version from Hilary, should we update that too?

.circleci/config.yml
```diff
@@ -1,13 +1,28 @@
...
+      - run:
+          name: Install Grunt
+          command: npm install --silent --global grunt grunt-cli
+      - run:
+          name: Install NPM dependencies
+          command: npm install --silent
+      - run:
+          name: Production build
+          command: grunt
+      # - run: grunt jshint
```
sallakarppinen: Why take out linting?

.circleci/config.yml
brecke: Well, hilary docker is still on ubuntu, isn't it? I remember I tried to migrate it to alpine at some point, but there was something blocking it, can't recall what exactly.

I don't think there is a problem though, because this is just to make sure 3akai-ux builds correctly so it's pretty agnostic.

brecke:
Hopefully when this gets merged, it will allow tests to pass for https://github.com/oaeproject/3akai-ux/pull/4191

PR #4198, opened 2018-03-02
Ability to join groups with moderation

ClaireLozano:
You can find the issue of the feature here : https://github.com/oaeproject/Hilary/issues/1402

---
This change is [Reviewable](https://reviewable.io/reviews/oaeproject/3akai-ux/4198)


brecke:
Here are the screenshots to give some context to my review comments:

screen shot 2018-03-23 at 15 46 14
screen shot 2018-03-23 at 15 49 38

ClaireLozano:
Review status: 0 of 20 files reviewed at latest revision, 19 unresolved discussions, some commit checks failed.

---

*[node_modules/oae-core/manageaccess/bundles/default.properties, line 3 at r1](https://reviewable.io/reviews/oaeproject/3akai-ux/4198#-L8fwUZ73siRlJG8WoKp-r1-3:-LCbev_k6kNNfgdplRWg:b-n9luxu) ([raw file](https://github.com/oaeproject/3akai-ux/blob/724087e74de4c662c2481afcb7afde8a7d4d49cb/node_modules/oae-core/manageaccess/bundles/default.properties#L3)):*
> Is this an error message?
Yes it is

---

*[ui/group.html, line 61 at r1](https://reviewable.io/reviews/oaeproject/3akai-ux/4198#-L8fwUZ73siRlJG8WoKz-r1-61:-LCbjz_qACz_uq7KhuJj:bcefoof) ([raw file](https://github.com/oaeproject/3akai-ux/blob/724087e74de4c662c2481afcb7afde8a7d4d49cb/ui/group.html#L61)):*
<details><summary><i>Previously, brecke (Miguel Laginha) wrote…</i></summary><blockquote>

I haven't tried this locally but I'm not sure this is right. Why do we need a new button exactly?

There is already a similar button [here](https://github.com/ClaireLozano/3akai-ux/blob/724087e74de4c662c2481afcb7afde8a7d4d49cb/ui/group.html#L80), why are there two?
</blockquote></details>

The button I created is to send a request when the join option is on 'apply to join the group' and the 'join group' button is to join the group when the join option is on 'anyone can join the group'.

---


*Comments from [Reviewable](https://reviewable.io/reviews/oaeproject/3akai-ux/4198)*

brecke:
Closed by https://github.com/oaeproject/3akai-ux/pull/4214

PR #4200, opened 2018-03-15
Fixes #4199

### LINKED ISSUE #4199: Update uglify to support ES2015+ syntax
That implies using `uglify-es` and `grunt-contrib-uglify-es`

brecke:
Requirejs optimize no longer uglifies
Now using uglify-es and grunt-contrib-uglify-es instead of requirejs

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -34,6 +34,29 @@ module.exports = function(grunt) {
         'qunit': {
             'files': ['tests/qunit/tests/*.html']
         },
+        'uglify': {
+            'my_target': {
+              'options': {
+                'mangle': {
+                  'reserved': ['jquery']
+                }
+              },
+              'files': [{
+                'expand': true,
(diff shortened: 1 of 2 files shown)


brecke:
we still have

```
'fileExclusionRegExp': /^(\.|<%= target %>|tests|tools|grunt|optimist|properties-parser|readdirp|underscore$|shelljs$|oae-release-tools|benchmark|robots\.txt)/,
```

where I'd try and remove these `grunt|optimist|properties-parser|readdirp|underscore$|shelljs$|oae-release-tools|benchmark` but I'm not sure that's a good idea. I just removed the ones that have been added during 2017 ±, I think I'll leave the others for another time, do you agree?

sallakarppinen:
Yeah, agreed. Grunt, optimist, oae-release-tools etc won't be used while the app is running anyway.

PR #4204, opened 2018-03-16
Documents in a folder should be shared with its group by default

ClaireLozano:
Issue : https://github.com/oaeproject/Hilary/issues/1407

---
This change is [Reviewable](https://reviewable.io/reviews/oaeproject/3akai-ux/4204)


brecke:
Is it ready then? to be reviewed again?

---

Review status: 0 of 1 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/oaeproject/3akai-ux/4204#-:-L8gxEtH9Fgs6vAhVeu3:brghh4h)*

node_modules/oae-core/setpermissions/js/setpermissions.js
```diff
@@ -115,12 +120,39 @@ define(['jquery', 'oae.core'], function($, oae) {
...
+                return callback(null, []);
+            }
+
+            oae.api.folder.getMembers(folderId, null, null, function(err, members) {
+                if (err) {
+                    return callback(err);
+                };
+
+                var membersProfile = _.map(members.results, function(member) { return member.profile; });
+                folderMembers = _.reject(membersProfile, function(profile) { return profile.id === oae.data.me.id; });
```
brecke: we could use `_.chain` here to make this simpler

node_modules/oae-core/setpermissions/js/setpermissions.js
```diff
@@ -81,19 +84,21 @@ define(['jquery', 'oae.core'], function($, oae) {
...
             });
 
+            selectedOutsideContext.concat(folderMembers);
+
+            // The item is always shared with the current context, so ensure it is high priority in
+            // the list of context names to display
+            if (pageContext.id !== oae.data.me.id) {
+                selectedContextNames.push(pageContext.displayName);
+            }
+
```
brecke: why are you changing the order the names are being displayed?

node_modules/oae-core/setpermissions/js/setpermissions.js
```diff
@@ -115,12 +120,39 @@ define(['jquery', 'oae.core'], function($, oae) {
...
+            oae.api.folder.getMembers(folderId, null, null, function(err, members) {
+                if (err) {
+                    return callback(err);
+                };
+
+                var membersProfile = _.map(members.results, function(member) { return member.profile; });
+                folderMembers = _.reject(membersProfile, function(profile) { return profile.id === oae.data.me.id; });
+
+                return callback();
```
brecke: instead of assigning a variable what is lost somewhere above, why not return it via callback? This way we avoid having "state" and we just pass arguments to functions accordingly.

node_modules/oae-core/setpermissions/js/setpermissions.js
```diff
@@ -141,8 +173,16 @@ define(['jquery', 'oae.core'], function($, oae) {
             // Trigger and bind an event that will ask for the page context
             $(document).on('oae.context.send.setpermissions.' + uid, function(ev, data) {
                 pageContext = data;
-                // Immediately pass the default permissions summary back to the widget caller
-                setUpAutoSuggest(savePermissionsChange);
+
+                // If the resource is inside of a folder, get the folder members to add them to the resource
+                getFolderMembers(pageContext.id, function(err) {
```
brecke: Let's have this function returning the members as part of the callback, and then pass it as an argument to `setUpAutoSuggest`. Nice and easy.

brecke:
Reviewed 1 of 1 files at r2.
Review status: all files reviewed at latest revision, 4 unresolved discussions.

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/oaeproject/3akai-ux/4204)*

node_modules/oae-core/setpermissions/js/setpermissions.js
ClaireLozano: It doesn't work when I do that :/ Should I let the code like this or should I push an unworking code to debug it together ?

brecke:
Reviewed 2 of 2 files at r3.
Review status: all files reviewed at latest revision, 4 unresolved discussions.

---

*[node_modules/oae-core/setpermissions/js/setpermissions.js, line 178 at r2](https://beta.reviewable.io/reviews/oaeproject/3akai-ux/4204#-L8fvqoPINOZqidnLZeg-r2-178:-L9L3Bdd4jbCM7QattSd:b-nxusod) ([raw file](https://github.com/oaeproject/3akai-ux/blob/f4dc97133085b59fd0291e9ce9fd3b5db1a2fb47/node_modules/oae-core/setpermissions/js/setpermissions.js#L178)):*
<details><summary><i>Previously, ClaireLozano (Claire Lozano) wrote…</i></summary><blockquote>

It doesn't work when I do that :/ Should I let the code like this or should I push an unworking code to debug it together ? 
</blockquote></details>

There's a third option here, which is to push working code which is also elegant :)

So, for `getFolderMembers` to return the members, that function needs to `return callback(null, member)` somewhere. If you fetch the members, as an array, then you'll be able to do this, somewhere. And then you can have 

```
getFolderMembers(pageContext.id, function(err, members) {´
if (err) {
...
}

// do something with members here
```

Does that make sense?

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/oaeproject/3akai-ux/4204)*

brecke:
Review status: all files reviewed at latest revision, 2 unresolved discussions.

---

*[node_modules/oae-core/setpermissions/js/setpermissions.js, line 143 at r2](https://beta.reviewable.io/reviews/oaeproject/3akai-ux/4204#-L8fvqoPINOZqidnLZeg-r2-143:-L9L54_WEKTR-f2_RHdd:bp8kfim) ([raw file](https://github.com/oaeproject/3akai-ux/blob/f4dc97133085b59fd0291e9ce9fd3b5db1a2fb47/node_modules/oae-core/setpermissions/js/setpermissions.js#L143)):*
<details><summary><i>Previously, brecke (Miguel Laginha) wrote…</i></summary><blockquote>

instead of assigning a variable what is lost somewhere above, why not return it via callback? This way we avoid having "state" and we just pass arguments to functions accordingly.
</blockquote></details>

This is still to be fixed. Let's use callbacks properly with `callback(null, data);`

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/oaeproject/3akai-ux/4204)*

ClaireLozano:
Oops I pushed the code :P 
It seems the code isn't executing the functions in the same order as in the previous commit ... or ... I don't know

brecke:
Ok, so.. 

There's this: you forgot to declare `folderMembers` [here](https://github.com/ClaireLozano/3akai-ux/blob/9008be494c97c78b784f94986cfdac9311f153ab/node_modules/oae-core/setpermissions/js/setpermissions.js#L137) but I think that somewhere you forgot to return a callback, because here:

```
oae.api.util.autoSuggest().setup($('#setpermissions-share', $rootel), {
                'allowEmail': true,
                'ghosts': widgetData.ghosts,
                'preFill': widgetData.preFill
            }, null, callback);
```

you're actually invoking `setUpAutoSuggest(folderMembers, savePermissionsChange(folderMembers));` and that is not a callback, I'm afraid. At least I think that's where the problem lies just by looking at it.

brecke:
:lgtm:

---

Reviewed 1 of 1 files at r5.
Review status: all files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/oaeproject/3akai-ux/4204#-:-L8gxYCJ9kj2KeOb87UP:bnfp4nl)*

ClaireLozano:
> (but I still have to test it locally, when Hilary is ready to review...)

But there is nothing to review on Hilary ... Maybe I've miss understanding you (?)
And why this comment doesn't appear on Reviewable ?

brecke:
I didn't make that comment on reviewable :) sorry!

Regardng Hilary, I'll check that tomorrow to be sure

ClaireLozano:
The redesign of the permission modal is done ;) ready to be reviewed ! 
There is only changes on the 3akai-ux project (forget the PR on Hilary)

brecke:
Closed. See https://github.com/oaeproject/Hilary/issues/1407

